PR #52, opened 2023-07-18
Spanish Translation

patofoto:
Have translated everything to spanish but cannot see context yet.


patofoto (commented):
Few grammar corrections

PR #82, opened 2023-08-20
add option to use measured or remaining weight (resolves #63)

### LINKED ISSUE #63: Measured weight option
When adding or checking used spools, I weigh them.
It would be nice to have the option to simply insert the measured results (in the edit and create screen of spools) and have the used weight variable update accordingly. 
We already know the weights of empty and full spools in the corresponding filament entry to do this calculation.

TomW1605:
i have only implemented this for the create page for now, decided to get feedback before porting it over to the edit page.

i tried a few ways to implement this with a system where whatever value you entered first became the one it used and the others get disabled and then auto calculate but i couldn't find a way to solve the circular references when you then delete the value you entered so you can switch to another one (code left in if you want to have a go with it). as a result i ended up adding some radio buttons so the user can select what field they want to use.

since the auto calculate relies on the filament weight i disabled all 3 fields till you select a spool. i also disabled the "measured weight" field if the filament doesn't list the spool weight but you can still use the "remaining weight" field.

there is still an error (`Warning: Cannot update a component ('InternalFormItem') while rendering a different component ('SpoolCreate').`) that shows up when you first select the spool but it shows up no matter how long you leave the page to render before selecting the spool and i don't have enough experience with react to work out what's wrong. however there also doesn't seem to be an actual problem caused by the error so im not sure if it "needs" fixing (i would still like to fix it).

lastly i removed the check in the api that prevents submitting both `remaining_weight` and `used_weight`, this is just temporary so i could test it and i assume we want to do some better check here. another option is to only ever submit `used_weight` no matter what weight they enter but i spent a while trying to work out how to do that and in the end decided this was quicker and works for the time being

### DIFF
--- a/client/public/locales/en/common.json
+++ b/client/public/locales/en/common.json
@@ -148,7 +148,7 @@
       "archived": "Archived"
     },
     "fields_help": {
-      "weight_to_use": "Select what weight value to enter. Measured weight is only avalable if the spool weight is set for the selected filament",
+      "weight_to_use": "Select what weight value to enter. Measured weight is only available if the spool weight is set for the selected filament",
       "used_weight": "How much filament has been used from the spool. A new spool should have 0g used.",
       "remaining_weight": "How much filament is left on the spool. For a new spool this should match the spool weight.",
       "measured_weight": "How much the filament and spool weigh.",
--- a/client/public/locales/en/common.json
+++ b/client/public/locales/en/common.json
@@ -133,8 +133,10 @@
       "filament_name": "Filament",
       "filament": "Filament",
       "material": "Material",
+      "weight_to_use": "Weight",
       "used_weight": "Used Weight",
       "remaining_weight": "Remaining Weight",
+      "measured_weight": "Measured Weight",
       "used_length": "Used Length",
       "remaining_length": "Remaining Length",
       "location": "Location",
@@ -146,7 +148,10 @@
       "archived": "Archived"
     },
     "fields_help": {
+      "weight_to_use": "Select what weight value to enter. Measured weight is only avalable if the spool weight is set for the selected filament",
       "used_weight": "How much filament has been used from the spool. A new spool should have 0g used.",
+      "remaining_weight": "How much filament is left on the spool. For a new spool this should match the spool weight.",
+      "measured_weight": "How much the filament and spool weigh.",
       "location": "Where the spool is located if you have multiple locations where you store your spools.",
       "lot_nr": "Manufacturer's lot number. Can be used to ensure a print has consistent color if multiple spools are used."
     },
--- a/client/src/pages/spools/create.tsx
+++ b/client/src/pages/spools/create.tsx
@@ -1,7 +1,7 @@
 import React, {useState} from "react";
 import {IResourceComponentsProps, useTranslate} from "@refinedev/core";
 import {Create, useForm, useSelect} from "@refinedev/antd";
-import {Form, Input, DatePicker, Select, InputNumber} from "antd";
+import {Form, Input, DatePicker, Select, InputNumber, Radio} from "antd";
 import dayjs from "dayjs";
 import TextArea from "antd/es/input/TextArea";
 import {IFilament} from "../filaments/model";
@@ -55,6 +55,7 @@ export const SpoolCreate: React.FC<IResourceComponentsProps & CreateOrCloneProps
   });
   filamentOptions?.sort((a, b) => a.label.localeCompare(b.label, undefined, {sensitivity: "base"}));
 
+  const [weightToEnter, setWeightToEnter] = useState(1);
   const [usedWeight, setUsedWeight] = useState(0);
 
   const selectedFilamentID = Form.useWatch('filament_id', form);
@@ -63,6 +64,21 @@ export const SpoolCreate: React.FC<IResourceComponentsProps & CreateOrCloneProps
   });
   const filamentWeight = selectedFilament?.weight || 0;
   const spoolWeight = selectedFilament?.spool_weight || 0;
+
+  const filamentChange = (newID: number) => {
+    const newSelectedFilament = filamentOptions?.find((obj) => {
+      return obj.value === newID;
+    });
+    const newFilamentWeight = newSelectedFilament?.weight || 0;
+    const newSpoolWeight = newSelectedFilament?.spool_weight || 0;
+
+    if (!(newFilamentWeight && newSpoolWeight)) {
+      setWeightToEnter(2);
+    }
+    if (!newFilamentWeight) {
+      setWeightToEnter(1);
+    }
+  };
 
   const weightChange = (weight: number) => {
     setUsedWeight(weight);
@@ -122,8 +138,12 @@ export const SpoolCreate: React.FC<IResourceComponentsProps & CreateOrCloneProps
             filterOption={(input, option) =>
               typeof option?.label === "string" && option?.label.toLowerCase().includes(input.toLowerCase())
             }
+            onChange={(value) => {
+              filamentChange(value);
+            }}
           />
         </Form.Item>
+
         <Form.Item
           hidden={true}
           name={["used_weight"]}
@@ -132,6 +152,20 @@ export const SpoolCreate: React.FC<IResourceComponentsProps & CreateOrCloneProps
           <InputNumber
             value={usedWeight}
           />
+        </Form.Item>
+
+        <Form.Item label={t("spool.fields.weight_to_use")} help={t("spool.fields_help.weight_to_use")}>
+          <Radio.Group
+            onChange={(value) => {
+              setWeightToEnter(value.target.value);
+            }}
+            defaultValue={1}
+            value={weightToEnter}
+          >
+            <Radio.Button value={1}>{t("spool.fields.used_weight")}</Radio.Button>
+            <Radio.Button value={2} disabled={!filamentWeight}>{t("spool.fields.remaining_weight")}</Radio.Button>
+            <Radio.Button value={3} disabled={!(filamentWeight && spoolWeight)}>{t("spool.fields.measured_weight")}</Radio.Button>
+          </Radio.Group>
         </Form.Item>
 
         <Form.Item
@@ -146,6 +180,7 @@ export const SpoolCreate: React.FC<IResourceComponentsProps & CreateOrCloneProps
             precision={1}
             formatter={numberFormatter}
             parser={numberParser}
+            disabled={weightToEnter != 1}
             value={usedWeight}
             onChange={(value) => {
               weightChange(value ?? 0);
@@ -164,7 +199,7 @@ export const SpoolCreate: React.FC<IResourceComponentsProps & CreateOrCloneProps
             precision={1}
             formatter={numberFormatter}
             parser={numberParser}
-            disabled={!filamentWeight}
+            disabled={weightToEnter != 2}
             value={filamentWeight ? filamentWeight - usedWeight : 0}
             onChange={(value) => {
               weightChange(filamentWeight - (value ?? 0));
@@ -183,7 +218,7 @@ export const SpoolCreate: React.FC<IResourceComponentsProps & CreateOrCloneProps
             precision={1}
             formatter={numberFormatter}
             parser={numberParser}
-            disabled={!(filamentWeight && spoolWeight)}
+            disabled={weightToEnter != 3}
             value={(filamentWeight && spoolWeight) ? filamentWeight - usedWeight + spoolWeight : 0}
             onChange={(value) => {
               weightChange(filamentWeight - ((value ?? 0) - spoolWeight));
--- a/client/src/pages/spools/create.tsx
+++ b/client/src/pages/spools/create.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import { IResourceComponentsProps, useTranslate } from "@refinedev/core";
 import { Create, useForm, useSelect } from "@refinedev/antd";
-import { Form, Input, DatePicker, Select, InputNumber } from "antd";
+import { Form, Input, DatePicker, Select, InputNumber, Radio } from "antd";
 import dayjs from "dayjs";
 import TextArea from "antd/es/input/TextArea";
 import { IFilament } from "../filaments/model";
@@ -15,7 +15,7 @@ interface CreateOrCloneProps {
 export const SpoolCreate: React.FC<IResourceComponentsProps & CreateOrCloneProps> = (props) => {
   const t = useTranslate();
 
-  const { formProps, saveButtonProps, formLoading } = useForm<ISpool>();
+  const { form, formProps, saveButtonProps, formLoading } = useForm<ISpool>();
 
   if (props.mode === "clone" && formProps.initialValues) {
     // Clear out the values that we don't want to clone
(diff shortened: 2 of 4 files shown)


Donkie:
Nice attempt :) I'm not entirely sure myself how this could be best implemented, but here are some points that I think will make it work better:

- Instead of disabling the unused inputs, prevent them from rendering entirely, e.g. `{weightMode == "used_weight" && <Form.Item blabla></Form.Item>`, that way, the un-rendered inputs value should never be sent to the server.
- If you do the above, you can just set the required attribute directly, no need to do the `rules` thing.
- Measured weight shouldn't be its own input sent to the server, it's too similar to remaining_weight for it to warrant additional logic in the server-side. Instead you should just calculate the correct remaining_weight and send that instead.
- The above point means you will have multiple fields with the name `remaining_weight` which complicates things with `setFieldValue`. I would probably create a react state to hold the computed `used_weight` value, attach that to a hidden input, then use that state as a central place to hold the weight value.
  - That way, we always only send `used_weight` to the server, through the hidden input
  - The currently active weight input's value is calculated using this central state
  - In the currently active weight input's onChange, you update the central state's value with the equivalent value from this input.
 
Let me know if you want further clarification!

TomW1605:
- i think its useful to have the calculated values at least displayed. that's why i went for disable rather than hide.
- im not sure exactly what you mean by this, do you mean i would just have 1 input that changes what it means instead of 3 where 2 are hidden?
- i agree there is no point sending all 3 values to the server. ideally i think we should only send one since the logic is already in place for calculating the other (and measured weight isn't handled at all)
- i agree this is a better way to go, it is what i wanted to do but not hide the other inputs and so have no hidden value for the central store. my idea was to always show all 3, then no matter what you picked to enter the other 2 would be calculated and then you would always send just `used_weight`. the problem i ran into was how to stop it sending the other 2 as well. 
  - i do think that sending an extra 2 numbers isn't really an issue and we can just ignore them on the server side and remove [line 171](https://github.com/Donkie/Spoolman/pull/82/files#diff-c3382e152f87cb88509e06dc42346284fde01220db49f530c47f9fd4f76c839cR171) since the spool object already handles only getting `used_weight`.

Donkie:
> * i think its useful to have the calculated values at least displayed. that's why i went for disable rather than hide.

That's a fair point, remove the name attribute from all the inputs then and use a hidden input to send the actual used_weight to the server.

> * im not sure exactly what you mean by this, do you mean i would just have 1 input that changes what it means instead of 3 where 2 are hidden?

You should have all 3 inputs then, and when the user changes value in any of them, they should just calculate a new value for the central state, and then the central state should update all 3 inputs with the correct value for them.

The central state is then sent to the server using a hidden input.

> the problem i ran into was how to stop it sending the other 2 as well.

As said above, try removing name from all of the inputs, and only send it using a hidden input

>   * i do think that sending an extra 2 numbers isn't really an issue and we can just ignore them on the server side and remove [line 171](https://github.com/Donkie/Spoolman/pull/82/files#diff-c3382e152f87cb88509e06dc42346284fde01220db49f530c47f9fd4f76c839cR171) since the spool object already handles only getting `used_weight`.

I don't want to remove any field from the server, or change how it works. That wouldn't be backwards compatible. You also can't send more than 1 field regarding weight, since then it becomes ambiguous for the server to know what field to choose from. So you have to send only a single field, preferably the used_weight field in this case.

TomW1605:
> > * i think its useful to have the calculated values at least displayed. that's why i went for disable rather than hide.
> 
> That's a fair point, remove the name attribute from all the inputs then and use a hidden input to send the actual used_weight to the server.

as i understand it, you still need the name attribute to reference the inputs in code, although i am not that familiar with react so i might be wrong about that. if there is a way then i thing this is a good option. although i do wonder why bother with the hidden field? why not just remove the name from remaining and measured and use the visible used field as the one sent to the server? it will be calculated from the other 2 exactly the same as the hidden one would be
 
> > * im not sure exactly what you mean by this, do you mean i would just have 1 input that changes what it means instead of 3 where 2 are hidden?
> 
> You should have all 3 inputs then, and when the user changes value in any of them, they should just calculate a new value for the central state, and then the central state should update all 3 inputs with the correct value for them.

i think this will run into the same circular logic issue i had before. unless there is some way to change the value through code that dosnt also trigger the on change function you need to disable all but 1 input

> > * i do think that sending an extra 2 numbers isn't really an issue and we can just ignore them on the server side and remove [line 171](https://github.com/Donkie/Spoolman/pull/82/files#diff-c3382e152f87cb88509e06dc42346284fde01220db49f530c47f9fd4f76c839cR171) since the spool object already handles only getting `used_weight`.
> 
> I don't want to remove any field from the server, or change how it works. That wouldn't be backwards compatible. You also can't send more than 1 field regarding weight, since then it becomes ambiguous for the server to know what field to choose from. So you have to send only a single field, preferably the used_weight field in this case.

ok thats fair. i looked and couldn't see anything using the remaining weight field for creating spools but i did forget it is an api that other things may use not just the spoolman front end

Donkie:
You can see how I've handled inputs together with states in `client\src\components\printing\printingDialog.tsx`, it's definitely possible. You wouldn't use setFieldValue then anymore, instead just manipulate the state using setState, and make the form take the value from that.

TomW1605:
cool, ill take a look

TomW1605:
i just thought, what do we do if the filament dosnt have a weight? should we be using remaining weight as out main value since that is more likely to be important and if you dont know the original weight of a spool you are not going to know how much was used. realistically the most likely weight someone will know is how much is left, and if that is all they know we cant calculate the used weight. how do you think we should handle that?

Donkie:
Remaining weight can only be used if the filament has a weight set. See the error produced in the create method in `spoolman/database/spool.py `

So if that's not the case, only used_weight should be available.

TomW1605:
i would propose that be changed, i think remaining weight is the most important weight. if i only know 1 weight about a spool of filament the one i want to know is how much is left.

that is a big thing to consider though so for now i will set it up so that if the filament dosnt have a weight value the remaining weight field is unavailable and if it dosnt have a filament weight and also dosnt have a spool weight the measured weight will be unavailable.

do you think these should be hidden or just disabled?

Donkie:
Probably disabled with a text explaining why.

The reason weight is required for remaining weight is that it's the used weight that is used for the actual counting. It's the basis for everything. That's because it technically is the simplest and most obvious thing, it starts at 0 and always counts up, and it can never go negative.

TomW1605:
yeah that makes sense, i was thinking more from a user perspective and didnt consider the always up and never -ve thing

TomW1605:
ok, i have made the changes and think i got it working how you want.

i haven't added in the message explaining why they are disabled but its 2:40am so that can wait till the morning. for now this is functional

client/src/pages/spools/create.tsx
```diff
@@ -105,17 +113,83 @@ export const SpoolCreate: React.FC<IResourceComponentsProps & CreateOrCloneProps
...
+            precision={1}
+            formatter={numberFormatter}
+            parser={numberParser}
+            value={used_weight}
+            onChange={(value) => {
+              set_used_weight(value ?? 0);
+              form.setFieldsValue(
+              {
+                used_weight: value ?? 0
+              });
```
Donkie, line 144: Create a function in here that calls both set_used_weight and form.setFieldsValue so that you don't have to calculate the value twice.

client/src/pages/spools/create.tsx
```diff
@@ -49,10 +50,17 @@ export const SpoolCreate: React.FC<IResourceComponentsProps & CreateOrCloneProps
...
+	    weight: item.weight,
+	    spool_weight: item.spool_weight
     };
   });
   filamentOptions?.sort((a, b) => a.label.localeCompare(b.label, undefined, { sensitivity: "base" }));
 
+  const [used_weight, set_used_weight] = useSavedState("create-used_weight", 0);
+
+  const selected_filament_id = Form.useWatch('filament_id', form);
+  const selected_filament = filamentOptions?.find(obj => {return obj.value === selected_filament_id});
```
Donkie, line 62: The naming convention in the typescript is lowerCamelCase, snake_case is only for the python part.

client/src/pages/spools/create.tsx
```diff
@@ -49,10 +50,17 @@ export const SpoolCreate: React.FC<IResourceComponentsProps & CreateOrCloneProps
     return {
       label: label,
       value: item.id,
+	    weight: item.weight,
+	    spool_weight: item.spool_weight
     };
   });
   filamentOptions?.sort((a, b) => a.label.localeCompare(b.label, undefined, { sensitivity: "base" }));
 
+  const [used_weight, set_used_weight] = useSavedState("create-used_weight", 0);
```
Donkie, line 59: This should be a regular react useState. `useSavedState` is a special version of it that I created in which the browser remembers the previous state, which is not something we want here.

client/src/pages/spools/create.tsx
```diff
@@ -49,10 +50,17 @@ export const SpoolCreate: React.FC<IResourceComponentsProps & CreateOrCloneProps
     return {
       label: label,
       value: item.id,
+	    weight: item.weight,
+	    spool_weight: item.spool_weight
     };
```
Donkie, line 55: Please look into running Prettier to format the code before you commit it. It will automatically indent the lines appropriately etc so everything looks neat. If you use VS Code then that is very simple to setup.

spoolman/api/v1/spool.py
```diff
@@ -159,10 +159,10 @@ async def create(  # noqa: ANN201
...
 ):
     if body.remaining_weight is not None and body.used_weight is not None:
-        return JSONResponse(
-            status_code=400,
-            content={"message": "Only specify either remaining_weight or used_weight."},
-        )
+       return JSONResponse(
+           status_code=400,
+           content={"message": "Only specify either remaining_weight or used_weight."},
+       )
```
Donkie, line 165: Undo this change

client/src/pages/spools/create.tsx
```diff
@@ -105,17 +113,83 @@ export const SpoolCreate: React.FC<IResourceComponentsProps & CreateOrCloneProps
...
+          initialValue={0}
+        >
+          <InputNumber
+            min={0}
+            addonAfter="g"
+            precision={1}
+            formatter={numberFormatter}
+            parser={numberParser}
+            disabled={!((selected_filament?.weight || 0) && (selected_filament?.spool_weight || 0))}
+            value={((selected_filament?.weight || 0) && (selected_filament?.spool_weight || 0)) ? (selected_filament?.weight || 0) - used_weight + (selected_filament?.spool_weight || 0) : 0}
```
Donkie, line 184: This is quite hard to read, can you perhaps split it up into a few variables? e.g. `const filament_weight = selected_filament?.weight || 0`

Donkie (changes requested):

client/src/pages/spools/create.tsx
TomW1605: i dont use vs code. my ide defaults to tab = 4 spaces, ill change that and reformat the code i have added

spoolman/api/v1/spool.py
TomW1605: sorry, i thought i did, apparently i removed the space as well as the commend

client/src/pages/spools/create.tsx
TomW1605: ok i will, im a bit confused as to now veriables work in react, putting const in front of things that change feels wrong so i keep forgetting that you can here (as long as you dont have to change them in code later)

client/src/pages/spools/create.tsx
Donkie: You can install prettier manually also and run it from the command line if you dont use vs code

client/src/pages/spools/create.tsx
Donkie: Yeah all you need to do is remove the first string argument and it works the same way.

client/src/pages/spools/create.tsx
Donkie: This whole function re-runs everytime react detects that something has changed which warrants a re-render, such as the value of a state has changed, or your filament watch is triggered. The const variables are only constant for a single specific render.

TomW1605:
ok, this commit should address all your comments

client/src/pages/spools/create.tsx
```diff
@@ -27,7 +27,7 @@ export const SpoolCreate: React.FC<IResourceComponentsProps & CreateOrCloneProps
     formProps.initialValues.filament_id = formProps.initialValues.filament.id;
   }
 
-  const { queryResult } = useSelect<IFilament>({
+  const {queryResult} = useSelect<IFilament>({
```
Donkie, line 30: Your prettier is making changes like this all over the place. Is it configured correctly?

Donkie (changes requested):

Donkie:
Function and code-wise it's quite good now!

However, UX-wise, I think it can be improved. The way you have listed the different inputs just next to each other like all the other inputs in the form, leads me to believe I need to fill all 3 of them in, and not that I should choose one of them. Normally you'd have some kind of radio button or a dropdown selector to first decide how you want to input your data. What do you think of this?

(if you decide to go this route, useSavedState for the input option is a perfect fit, so you don't have to change the input method every time you open the form :) )

TomW1605:
> Normally you'd have some kind of radio button or a dropdown selector to first decide how you want to input your data. What do you think of this?

so should i go back to what i had before UX wise but use the current system for calculations, display update and sending to the server?

Donkie:
> > Normally you'd have some kind of radio button or a dropdown selector to first decide how you want to input your data. What do you think of this?
> 
> so should i go back to what i had before UX wise but use the current system for calculations, display update and sending to the server?

Yeah perhaps that's a good idea, it's up to you, see what looks good. I just think it's a bit confusing as it is now.

client/src/pages/spools/create.tsx
TomW1605: ok, i cant work out how to configure it properly. there is a `.prettierrc` file but it dosnt have anything about spaces inside braces. could you check that the `.prettierrc` file in the repo is up to date with the config your using?

TomW1605:
ok, this should work well now. i decided not to use useSavedState for this because the options that are available depend on the filament that is selected. since we need to calculate used weight if you dont provide a spool weight you cant enter the measured weight and get a useful used weight. and since when you make a new spool it starts with no filament selected it needs to default to assuming that there is no filament or spool weight.

Donkie:
This looks great! Seems to work as intended as well, I think you can go ahead and make it for the edit screen now as well.

TomW1605:
will do. do you think we need any special handling for duplicate or is defaulting to used weight ok?

Donkie:
Not sure what you mean by duplicate

TomW1605:
when you duplicate a spool you get the create page but with some values pre filled. is there anything that needs to change with that for this feature? for example should it default to the same weight type you used for the spool your duplicating from?

Donkie:
Nah that I don't think is necessary. Just fill in the computed values and let the client decide how they want to enter it

TomW1605:
ok, added the options to the edit screen. that should be it

client/public/locales/en/common.json
```diff
@@ -146,7 +148,10 @@
       "archived": "Archived"
     },
     "fields_help": {
+      "weight_to_use": "Select what weight value to enter. Measured weight is only avalable if the spool weight is set for the selected filament",
```
Donkie, line 151: Typo, "available", also missing period at the end.

Donkie (changes requested):

Donkie:
Just one minor comment and then this is ready for merge :)
Sorry for the dragged out process, I don't have much time except for weekends to work on this.

client/src/pages/spools/edit.tsx
```diff
@@ -35,14 +35,60 @@ export const SpoolEdit: React.FC<IResourceComponentsProps> = () => {
...
     formProps.initialValues["filament_id"] = formProps.initialValues["filament"].id;
   }
 
+  useEffect(() => {
+    if (formProps.initialValues && usedWeight != formProps.initialValues["used_weight"]) {
+      setUsedWeight(formProps.initialValues["used_weight"])
+    }
+  }, []);
+
+  console.log(usedWeight)
```
Donkie, line 90: Remove

TomW1605: just the console.log right? not the useEffect part?

Donkie: Yeah correct

TomW1605:
ok both should be fixed

PR #94, opened 2023-09-01
Docker: set permissions for config directory

mkuf:
Hi there, 

while adding spoolman to [prind](https://github.com/mkuf/prind), I noticed that when using a named volume to store the spoolman-db, permissions for `/home/app/.local/share/spoolman` are set to `root:root` which causes the server to fail as the `app` user lacks permissions to write to it. 

Example: 
```yaml
services:
  spoolman:
    image: ghcr.io/donkie/spoolman:edge
    restart: unless-stopped
    volumes:
      - spoolman-db:/home/app/.local/share/spoolman

volumes:
  spoolman-db:
```

This PR creates the database directory and sets `app:app` as owner at image build time.
Furthermore, to follow [Best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#volume) adds the `VOLUME` directive to the dockerfile for the database directory.

Looking forward to your response.
-Markus

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,10 @@ LABEL org.opencontainers.image.licenses=MIT
 RUN apk add --no-cache libstdc++
 
 # Add local user so we don't run as root
-RUN adduser -D app
+RUN adduser -D app \
+ && mkdir -p /home/app/.local/share/spoolman \
+ && chown -R app:app /home/app/.local/share/spoolman
+
 USER app
 
 # Copy built client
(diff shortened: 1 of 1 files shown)


Donkie:
Seems to work well, thank you!

Donkie:
How important is the VOLUME directive for you @mkuf ?

I've been looking into the issue people are having with data going missing etc. One issue is that they forget to mount the data dir, or they mount it incorrectly. If I remove the VOLUME directive from the Dockerfile, I will be able to inside the container detect if the directory is mounted or not (that info is available from the unix `mount` command). However if the VOLUME directive is there then I can't distinguish if the data dir has been mounted by the user or just by the default anonymous volume that docker creates.

If I then can detect that it's not mounted, I can warn people that data will go poof in the server logs

mkuf:
There is no requirement for the volume directive, its just best practice. 
If it makes life easier for you and users of spoolman, you can remove it from the image without breaking anything. 

The important thing about this PR is that the directory is created and permissions for it are set at build time, so named docker volumes can be used. 

-Markus

PR #107, opened 2023-09-13
Add Czech localization

MickCZ:
Spoolman/client/public/locales/cs


Donkie:
Your file setup is a bit wrong. The common.json in the parent folder shouldn't be there. You're also missing the i18n.ts addition. Please see for example #77 for reference.

MickCZ:
file i18n.ts completion

Donkie:
You still have one common.json too many.

PR #141, opened 2023-10-09
Install script

mlee12382:
I have included my tested and working install script to install Docker and create and launch the container for Spoolman along with adding a readme for the installer. The install git clone currently points to my fork but that's easily changed or you can just add the instructions to your main readme.


Donkie:
I appreciate your contribution! I think we should make it more clear however that this script is purely aimed at raspberry pis.

I think we should put the script inside a /scripts folder. I imagine we will add more in the future.

The readme should be shortened and added to the main README instead. Most of the text you added in your readme there should be added as a disclaimer in the script that is shown when it's run, and the user needs to press Y to continue.

You should also add a check in the script to make sure people don't run it as sudo. I imagine that would break the script if people did.

Also what timezone trickery are you doing at line 54?

mlee12382:
> I appreciate your contribution! I think we should make it more clear however that this script is purely aimed at raspberry pis.
>
I think it should work on other Linux devices also? I did build it with the thought that it would most likely be used on the same system that Klipper is running on though. It should still grab the home directory for whatever the system is using at least in theory. 

> I think we should put the script inside a /scripts folder. I imagine we will add more in the future.
> 
> The readme should be shortened and added to the main README instead. Most of the text you added in your readme there should be added as a disclaimer in the script that is shown when it's run, and the user needs to press Y to continue.
>
I can do that, I was also thinking about adding a reboot prompt after docker is installed, and the permissions are set since the permissions don't seem to work until you reboot . I have the warning at the end, but I think I should just set it up to prompt for the reboot instead. 

> You should also add a check in the script to make sure people don't run it as sudo. I imagine that would break the script if people did.
> 
Doesn't it need sudo for installing docker? I will test if running it as sudo will break it when I work on the reboot revision. I can definitely add a check for that if needed, though.

> Also what timezone trickery are you doing at line 54?

It is checking the system timezone and inserting it in the container.

Donkie:
> I think it should work on other Linux devices also? I did build it with the thought that it would most likely be used on the same system that Klipper is running on though. It should still grab the home directory for whatever the system is using at least in theory.

You're using the raspbian version of docker (see line 18).

> I can do that, I was also thinking about adding a reboot prompt after docker is installed, and the permissions are set since the permissions don't seem to work until you reboot . I have the warning at the end, but I think I should just set it up to prompt for the reboot instead.

I have never needed a reboot for these things in unix. I think a warning is good enough.

> Doesn't it need sudo for installing docker? I will test if running it as sudo will break it when I work on the reboot revision. I can definitely add a check for that if needed, though.

Yes it does, but you're already using sudo inside the script, so that solves that issue. What you need to prevent is people doing `sudo ./install_spoolman.sh`, which will cause things like home directory etc inside the script to become wrong.

> It is checking the system timezone and inserting it in the container.

That is not what you're doing from my understanding. You're already doing that at line 51. The line 54-57 stuff doesn't seem to be doing anything useful.

mlee12382:
I'm still learning a lot of this stuff, I had ChatGPT helping me write the script, haha. 

I think I can remove the gpg key and apt repository, I forgot those were in there from trying to get it working that way. It's using the curl line on 22 now instead of apt.

I wasn't sure if you needed to have sudo access for the script to use sudo commands. Does it just prompt for a su password if needed by the script then?

Not sure about line 57 but 54-56 just display that the script is setting the tz to the system setting and echoing the tz being used. At least the best I can tell.

mlee12382:
I have updated the script to include the reboot if needed and to give a link to the rest of the documentation for adding it to moonraker.

mlee12382:
It appears there is an error with the tests for arm64, any ideas what that is about? Thoughts on how to fix it?

install_spoolman.sh
```diff
@@ -0,0 +1,96 @@
...
+        target: /home/app/.local/share/spoolman
+    ports:
+      - "7912:8000"
+    environment:
+      - TZ=$(cat /etc/timezone) # Set timezone based on system timezone
+EOF
+
+# Change timezone to match the system timezone
+echo "Changing timezone to match the system timezone..."
+echo $(cat /etc/timezone) | sudo tee /etc/timezone
```
rufinus, line 86: thats makes no sense. you read /etc/timezone to write to /etc/timezone? what you do in line 81 is better and more then enough.

mlee12382: Yep, I think I mentioned I was using chatgpt to help write the script and help with testing. For some reason it misunderstood that I only wanted to pull the system tz for the container. I had tried to get it to explain why it was there and it said it was needed but I finally got it to understand that I didn't need it there.
I am still learning how to use this programming language and it's been 15yrs since I've really done much programming in general so I'm a bit rusty with what I do know lol. It's removed now.

mlee12382:
@Donkie I have renamed and moved the script into the script folder and I also added a github backup script along with instructions to keep the database backed up.

Donkie:
I appreciate your effort, but I honestly think that we shouldn't provide a script to automatically install docker for people. It's quite invasive, and docker themselves say that the get.docker script shouldn't be used for production environments. It's better if we just point people to the official docker installation guide (which is what the README already does) so they will get the correct docker version for their OS. I believe that if people aren't able to setup docker on their own on their linux machine with the 1000s of guides available online, then they're probably in over their heads to begin with. Additionally, we now have an officially supported standalone installation, so an automatic script like this is less necessary.

The 2nd half of your script where you setup the docker-compose etc, is a nice feature though. Should prevent issues with people misconfiguring their docker compose.

Regarding the automatic backup script, it's an interesting feature, but it may be more suitable for a discussion post rather than being built directly into Spoolman. Backup preferences can be quite specific, and using a GitHub repository for this purpose is not very common.

mlee12382:
I have removed the Docker install portion of the script so it's only setting up the container and launching it. I agree I think more people will be using the standalone installer unless they're already using Docker.

My thought with the backup script is that it would be an optional install but available with instructions for those that wanted to have an offsite backup of their database. The github backup process is nice because it's fairly easy to set up and specifically in relation to klipper it's a great way to keep track of all of your changes over time, though that seems less useful in the case of Spoolman. The process is detailed pretty well in the Voron Docs so for those that have already set it up for that or want to it makes sense to use the same method imo.

Donkie:
You're still meddling a bit with system level things, by adding the user to the docker group. Just adding someone to the docker group doesn't guarantee that they will be able to run docker commands without sudo, more things are needed to do that. I think it's easier if you just prefix the docker commands with sudo instead of messing with the user group.

I still don't really like the backup script approach. It's easy enough for the few people who want to use it to add this themselves by following a guide in discussions.

PR #159, opened 2023-10-20
feat: add basic-auth mechanism

slaesh:
- add the ability to add basic-auth mechanism
- will protect both, UI and API
- preparations done, to add API-tokens later on

PS: I didnt do that much python before.. so please feel free to adjust whatever you want or just be super picky and tell me what to do :))

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,13 @@ If you want to connect with an external database instead, specify the `SPOOLMAN_
 | SPOOLMAN_LOGGING_LEVEL    | Logging level, any of: `CRITICAL`, `ERROR`, `WARNING`, `INFO`, `DEBUG`, defaults to `INFO`.                                  |
 | SPOOLMAN_AUTOMATIC_BACKUP | Automatic nightly DB backups for SQLite databases. Enabled by default, set to `FALSE` to disable.                            |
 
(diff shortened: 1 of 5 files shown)


slaesh:
todo:

- [ ] update docs

Donkie:
While I appreciate your contribution, I'm not sure that this is the way I want to go with authentication for Spoolman. Making authentication correctly is not simple, and there are a few issues with this approach:
* This is hard to extend with more functionality. What if we want multiple accounts with different levels of privileges? What if we want other authentication methods, like API key? We'd have to get rid of the new environment variables you created, which would break backwards compatibility.
* There are no measures in place to prevent abuse, like brute forcing the login details.

If we are to implement an authentication system to Spoolman I think it should be done in a robust and extensible way from the beginning, with the ability to setup additional accounts, create api keys on the fly, etc.

For now I think it's better to let already existing solutions provide an authentication layer. Authelia can do this for example. I have been planning to create a guide on how to setup Authelia together with Spoolman.

slaesh:
Yeah that would be great. This is just a quit and dirty way, for sure. was trying to host spoolman and was afraid that there is no authentication needed. :))
so better a bad one, than none? ;)
but I get the point! unfortunately this is not "my" software stack at all, so I can't really help.. sorry!
so I will life with this fork for the moment until there is something better build in 🙏😎😊

PR #164, opened 2023-10-22
Add support for sharing table views via link

clevelandcs:
What does this do?
* Adds share button to top of list view
  * Share button shares current table sorts, filters, pagination, and columns with hash parameters
  * Sharing is done via link with 
    * a) [web share api](https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API) or 
    * b) updates link hash in URL bar for user to copy if share API is not supported by browser
* Automatically sets table sorts, filters, pagination, and columns on load when visiting Spoolman with full share link

What is the use case for this feature?
* Enables sharing of views in multi-user environments via simple link
* Adds support for simple and quick location and/or material based inventory counts - in my case writing the link for the shelf or drybox location filter to NFC tags and QR codes attached to the shelf.

What is missing?
* Possible translation issues. I'm not a linguist but did my best to add translations for the share button where applicable
* Testing, I spent about an hour testing different scenarios but it would be nice to have more than one pair of eyes
* QR code generation could possibly be added using this type of link for shelf ID labels

Please let me know if you see problems or things you'd like changed to get this merged. This is my first open-source contribution and I'm not a React developer (Vue+.NET is my wheelhouse). Thanks for the work you've done on this great tool!

### DIFF
--- a/client/src/utils/saveload.ts
+++ b/client/src/utils/saveload.ts
@@ -1,7 +1,6 @@
 import React from "react";
 import { CrudFilter, CrudSort } from "@refinedev/core";
 import { isLocalStorageAvailable } from "./support";
-
 interface Pagination {
   current: number;
   pageSize: number;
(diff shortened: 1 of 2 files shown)


Donkie:
Thank you for your contribution! I agree this is something that is definitely needed.

Isn't it better though if we make the current window location reflect the filter/sorting/etc state instead of having it be generated by a button? That is, remove the button and instead update the browser window location on every render or something. Refine has some built-in functionality for this, the syncWithLocation variable on the useTable hook, but I have deliberately disabled that because it isn't fully compatible with the current way of doing filters and such.

Perhaps your method can be used instead, and we just have to figure out how to set the window hash at appropriate times.

clevelandcs:
I do agree with the idea of having the current window location reflect the current state of the table. The only reason I did it this way was concern for the perceived uglyness with the length of the location hash when it contains the full state of the table.

Perhaps it would be ideal to have the table state only as it differs from the initial state shown in the location hash. I will modify the behavior to that and see how it works. I'll update the PR with the changes of add another comment here if I come up with something else.

PR #168, opened 2023-10-23
Add support for sharing table views via link

clevelandcs:
Updated PR changing UX to follow discussion in #164 
What does this do?
* Automatically sets table sorts, filters, pagination, and columns on load when visiting Spoolman with full share link
* Updates URL hash fragment to reflect current state of table 
* ~~Adds share button~~ 
* Uses ~~share API or~~ current URL to provide sharable link ~~when clicking share~~

What is the use case for this feature?
* Enables sharing of views in multi-user environments via simple link
* Adds support for simple and quick location and/or material based inventory counts - in my case writing the link for the shelf or drybox location filter to NFC tags and QR codes attached to the shelf.

What is missing?
* ~~Possible translation issues. I'm not a linguist but did my best to add translations for the share button where applicable~~
* Testing, I spent about an hour testing different scenarios but it would be nice to have more than one pair of eyes
* QR code generation could possibly be added using this type of link for shelf ID labels
* In the future Refine's syncWithLocation could potentially support customization (I could not find a good way to do this) so that state is reflected in URL without this extra logic

Please let me know if you see problems or things you'd like changed to get this merged. This is my first open-source contribution and I'm not a React developer (Vue+.NET is my wheelhouse). Thanks for the work you've done on this great tool!

### DIFF
--- a/client/src/utils/saveload.ts
+++ b/client/src/utils/saveload.ts
@@ -1,7 +1,6 @@
 import React from "react";
 import { CrudFilter, CrudSort } from "@refinedev/core";
 import { isLocalStorageAvailable } from "./support";
-
 interface Pagination {
   current: number;
   pageSize: number;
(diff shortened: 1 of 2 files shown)


client/src/utils/saveload.ts
```diff
@@ -35,28 +34,44 @@ export function useStoreInitialState(tableId: string, state: TableState) {
     if (isLocalStorageAvailable) {
       localStorage.setItem(`${tableId}-sorters`, JSON.stringify(state.sorters));
     }
+    if (JSON.stringify(state.sorters) != JSON.stringify([{ field: "id", order: "asc" }])) {
+      updateURLHash(`${tableId}-sorters`, JSON.stringify(state.sorters));
```
Donkie, line 38: I'm wondering if it's necessary to include the tableId in the hash, since there will only be a single table per page anyway. The page URL itself will kinda decide the tableId.

Donkie:
This seems very nice, I like that you managed to make it all without having to touch the components themselves. 
I have tried it and the functionality seems to work well. One thing I noticed is that it doesn't seem to update when I click the "Clear Filters" button.

I'm also thinking if showColumns should be part of the hash. Is that really something you want to share?

I'm also wondering if perhaps the hash could be cleaned up a bit. This actually goes for the data saved in the localStorage as well. Sometimes it stores an empty filter, so perhaps we should filter out empty filters before saving and adding to the hash?

client/src/utils/saveload.ts
clevelandcs: You're probably right about that. I was just keeping the keys the same between localStorage and the hash. Excluding it would shorten the hash a little bit and it should be unnecessary.

clevelandcs:
I've changed the table setting updater functions as discussed. Both localStorage and the URL hash are only being set if the change differs from the default table behavior. In addition, in the case of filters only filters that actually contain a filter value are being saved this significantly reduces the number of characters that needed to be saved.

Donkie:
I think it looks good now, so I'll go ahead and merge it. Thanks for the contribution again!

Issue #101, opened 2023-09-09
Set default Value in Create Filament and "Save and Continue" Option


TjardF:
Like Title said:
I'd set a default of 0m (which can be overwritten when inputting new number without deleting the 0)
Also I'd add a Button to directly add the next Role.

https://refine.dev/docs/advanced-tutorials/forms/save-and-continue/#handling-submit-events-on-buttons

Just a reminder for me to try to add those things and create PR, if nobody is still doin this...

Donkie:
What fields do you want to have default 0?

TjardF:
The used length of a spool, cause its set to no default& not null.
Also a add next reel button would be good. At least for the first time to add all the spools at once

Donkie:
The add flow was rebuilt from scratch in 0.26.0 and is much quicker for entering several spools in a row. Closing this one as stale against the old client, if you still want configurable field defaults please open a fresh issue against the new client.

Issue #76, opened 2023-08-15
Easy way to jump from spool show view to filament show view


gpmfuchs:
1. When opening a spool show window, it would be really nice to have a link (and maybe the data) to the corresponding filament visible (to click on).

2. When opening a filament show window, it would be nice to have a list of all the spools of this filament at the end of the page.
Or to have a link which jumps to a filtered view of the spools windows which only shows the spools of this filament.

Thanks for the good work :-)

Donkie:
1. That is already added in 0.11.0, make sure you've updated and cleared your browser cache (Ctrl+F5)
2. Good idea :)

Thank you!

Donkie:
The new Library view in 0.26.0 combines manufacturers, filaments and spools into one tree, so a filament's spools sit directly under it. Closing as covered by the new client.

PR #172, opened 2023-10-25
#76 spool list, no reload on page change

clevelandcs:
* Completes requested feature from #76 adding filtered route link to end of record as "Spool List" alternate name suggestions "Show Spools", "List Spools", "View Spools", etc 
* Replaces href link for show Vendor with React router navigate function to avoid unnecessary reload of the SPA
* Requested feature #101  Save and continue button on Spool Create- added to all 3 resources so footer buttons are consistent across app
* Add input to the footer for adding an arbitrary number of spools of the same filament on Spool Create - reduces the tedium when receiving a shipment of 20 spools of the same black
* Fix regression of filter URL containing empty filters due to using state filters instead of filtered filters list


client/src/utils/overrides.css
```diff
@@ -0,0 +1,7 @@
+.ant-input-number-group-addon {
+    padding: 0 !important;
+}
+
+.ant-input-number-input {
+    text-align: center !important;
+}
```
Donkie, line 7: These overrides makes the quantity selector look good indeed, but it makes all the other number inputs look quite wonky...

client/public/locales/en/common.json
```diff
@@ -1,263 +1,265 @@
 {
-    "actions": {
```
Donkie, line 2: See if you can re-save this file without messing up the entire formatting

Donkie:
Very nice additions, thank you! Just a few minor comments to address before merge.

client/public/locales/en/common.json
clevelandcs: Just pushed, inconveniently I had to do it with nano because VSCode insists on formatting on save.

client/src/utils/overrides.css
clevelandcs: I was able to do the text-align using CSS ids (unfortunately it doesn't seem to allow !important on inline styles). I had to switch away from using the input addons because they don't accept ids. It doesn't look exactly as nice as it did in this commit but it does look similar and avoids messing other inputs.

client/public/locales/en/common.json
Donkie: I format using prettier in VSCode and it works fine for me. Is it using the prettier config in the client directory properly?

Donkie:
Found another issue after some more testing, I get the "are you sure you want to leave and lose your changes?" popup when I click the regular save button now on the spool create page after I've set the quantity to more than 1.

clevelandcs:
> Found another issue after some more testing, I get the "are you sure you want to leave and lose your changes?" popup when I click the regular save button now on the spool create page after I've set the quantity to more than 1.

Hmm, that seems like it could be a bit difficult to work around. I don't see anything builtin to disable that. Just a quick search seemed to suggest I'll need to add a bunch of js to override the default behavior to ignore just that one field.

clevelandcs:
> > Found another issue after some more testing, I get the "are you sure you want to leave and lose your changes?" popup when I click the regular save button now on the spool create page after I've set the quantity to more than 1.
> 
> Hmm, that seems like it could be a bit difficult to work around. I don't see anything builtin to disable that. Just a quick search seemed to suggest I'll need to add a bunch of js to override the default behavior to ignore just that one field.

Update: Ok, there's simply a "warnWhenUnsavedChanges" property on useForm that can be set to false. Setting this to false disables the default unsaved warning seems like this is a fine way forward at least as far as the negligible data entry that could need to be reentered if someone accidentally did refresh or navigate away from the page.

Donkie:
sounds good to me!

PR #185, opened 2023-11-03
docs: added reference to home assistant integration

Disane87:
I added a reference to my home assistant integration. I guess many of Spoolman users could benefit from it 😸


Donkie:
Good idea :)

PR #197, opened 2023-11-12
(docs) add initial list of FAQs

mkevenaar:
Hi,

I noticed that a lot of people create issues with the camera not being able to be accessed over HTTP. Therefore I started a FAQ.

A FAQ with only one question seemed a bit useless so I added a translation question.

Let me know if you want any changes.


Donkie:
Good idea, thank you! :)

PR #211, opened 2023-12-01
Update README.md

Travis90x:


Donkie:
Your changes to the standalone installation doesn't really add any value. It's also incorrect, since a recent issue showed that you need to CD into the spoolman folder (like it also already says in the README). You also removed step 2.

The moonraker configuration part is not supposed to be here, it clutters it up too much.

Donkie:
I have updated the README myself now, so I'll close this.

Issue #88, opened 2023-08-25
Add a price for each spool


sudermanjr:
I often buy spools at wholesale/bulk prices and then buy one or two more later at MSRP.

It would be really nice to have the filament price be my MSRP, and then be able to put in the price I paid for each spool.

This is something I would be willing to contribute to the backend if it's a feature folks want, but I'm terrible at JS so the Frontend may be out for me

P.S. Thanks for this project, it's exactly what I needed.

Donkie:
Yes, a price field on spools is something others have suggested as well. Definitely something worth adding!

ThomasHineXYZ:
This would definitely be appreciated.
Filament spool prices aren't always consistent. On top of that, there might be a sale or coupon or something.

pdsccode:
Adding my interest here as I just reordered a new spool of an already in my DB existing filament for a different price. 🙂

PR #240, opened 2023-12-30
Adding prices to spools

pdsccode:
As requested in https://github.com/Donkie/Spoolman/issues/88, prices for single spools are practical for instances where a new spool should be added that has been bought for a different price.


Donkie:
Thank you for your contribution!

I think it would be nice if the displayed price in the UI would default to the filament's price if the spool doesn't have any price set.
You should be able to mutate the price field in the `collapseSpool` function in list.tsx to achieve this.

Also please add integration tests for this in tests_integration/tests/spool

pdsccode:
Good idea with the default. Didn't think about this. Will add it and also the tests.

.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
 default_language_version:
-  python: python3.9
+  python: python3
```
Donkie, line 2: Undo this change

client/public/locales/en/common.json
```diff
@@ -150,6 +151,7 @@
             "archived": "Archived"
         },
         "fields_help": {
+            "price": "Price of a full spool in the system configured currency.",
```
Donkie, line 154: Here you should add that if this is not set, the filament's price will be assumed instead.

client/src/pages/spools/show.tsx
```diff
@@ -58,6 +58,8 @@ export const SpoolShow: React.FC<IResourceComponentsProps> = () => {
       <NumberField value={record?.id ?? ""} />
       <Title level={5}>{t("spool.fields.filament")}</Title>
       <TextField value={record ? filamentURL(record?.filament) : ""} />
+      <Title level={5}>{t("spool.fields.price")}</Title>
+      <NumberField value={record?.price ?? ""} />
```
Donkie, line 62: Here you should also default to the filament's price, to make it consistent with the list view

Donkie (changes requested):

Donkie:
Tested it locally now and it generally seems to work fine. I have some minor comments I want you to address before I merge it :)

.pre-commit-config.yaml
pdsccode: I changed it back now. Accidentally worked with 3.11 and got pre-commit not running without this change. Works now.

client/public/locales/en/common.json
pdsccode: Added as requested.

client/src/pages/spools/show.tsx
pdsccode: Filament price is now shown on spools/show.tsx when spool price is not set :)

Donkie:
Sorry for the delay, it's merged now. Thank you for the help! :)

pdsccode:
No worries. Thanks for letting me help and advise me as well. 😊

PR #278, opened 2024-01-23
Switch tests from "docker-compose" to "docker compose"

hcooper:
I was running some tests and ran into the error `kwargs_from_env() got an unexpected keyword argument 'ssl_version'`. [The suggestion](https://stackoverflow.com/questions/77641240/getting-docker-compose-typeerror-kwargs-from-env-got-an-unexpected-keyword-ar
) of just switching to the newer `docker compose` command worked great for me (all `pdm run itest` passes).

I'm hoping this should be a drop-in change with no impact.

### DIFF
--- a/tests_integration/run.py
+++ b/tests_integration/run.py
@@ -35,9 +35,9 @@ if __name__ == "__main__":
 
     for target in targets:
         print(f"Running integration tests against {target}...")
-        os.system(f"docker-compose -f tests_integration/docker-compose-{target}.yml down -v")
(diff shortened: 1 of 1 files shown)


rufinus:
in shell scripts i use this:


```
dc() {
    if ! command -v docker-compose &> /dev/null
    then
        docker compose "$@"
    else
        docker-compose "$@"
    fi
}

dc up -d
```

Donkie:
Good point, thanks for the update!

PR #279, opened 2024-01-23
Support dynamic ports

hcooper:
This add support for overriding the default port used by Spoolman's docker container. 8000 is still used unless the `SPOOLMAN_PORT` envvar is provided. Also `SPOOLMAN_HOST` can be used to change the address bound to, but defaults to `0.0.0.0`.

Motivation for this change is that not all docker architectures use port mapping (e.g. it's possible each container gets its own routable IP instead). In these cases services need to bind directly to the port number they're expected to be reachable on (e.g. port 80).

Other work done:
 - Updated docs
 - Updated integration tests to handle port overrides
 - Created new integration test specifically for port overrides (using an sqlite backend)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,4 +62,3 @@ ENV BUILD_DATE=${BUILD_DATE}
 # Run command
 EXPOSE 8000
 ENTRYPOINT ["/home/app/spoolman/entrypoint.sh"]
-CMD ["--host", "0.0.0.0", "--port", "8000"]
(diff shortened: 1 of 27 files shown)


Donkie:
Nice idea!

We already have env vars for host and port, used for the standalone install method. I just have forgotten to add them to the README.md. See https://github.com/Donkie/Spoolman/blob/master/scripts/start.sh for example. So use these variables instead.

Also, I don't think it's necessary to add the docker-compose-port integration test case. We don't need to integration test something like this, which should always work.

hcooper:
Merged with recent other changes. Removed the port test case. Think this should just work alongside the standalone envvars fine :-)

Donkie:
Seems to work, just one minor comment and then I'll merge it

hcooper:
Changed to `SPOOLMAN_PORT`as requested. I was following the pattern set by `DB_TYPE` which is `SPOOLMAN_DB_TYPE` in the spoolman jobs, and `DB_TYPE` in the tests. So that might want aligning too. e.g. https://github.com/Donkie/Spoolman/blob/master/tests_integration/docker-compose-postgres.yml#L27

Donkie:
good point, that can be fixed in the future, let's merge this for now

PR #311, opened 2024-02-09
Feature: Add option to use full URL in QR codes instead of `web+spoolman:` prefix

0x416E64:
I would like to have a QR code with a link I can open in any barcode scanning app, rather than having to open Spoolman first, then it's scanning view and then scan finally scan the QR code. Also with QR codes generated with this options a person without knowledge of the Spoolman URL can find spool info.

Quickly bodged together, tested both from Spoolman & barcode scanning app, uses the current browser origin for the link generation which could be confusing if the user accesses the instance using multiple origins (`localhost`, lan IP, domain name), but otherwise the user would have to specify the base URL.

Also there is an edge case with the in-app scanner if a person uses more than 1 instance of Spoolman (let's call them Spoolman A, Spoolman B), but I don't know if that's worth solving (and how to solve it easily):

1. Generate a QR code in Spoolman A for spool ID 1 -> links to https://a.example.com/spool/show/1
2. Scan with the in-app scanner from Spoolman B -> the regex match doesn't check that the link has the same origin -> navigates to https://b.example.com/spool/show/1, showing a different spool

This behavior could be possibly desired to support the multiple access origin case from above, also it's the same behavior as the current web+spoolman: prefix.

I'm open to any suggestions, extra changes or corrections. Also this option should probably be better documented/explained somewhere else than this PR and the one sentence description next to the toggle switch, but i didn't see any place to put the documentation.


Donkie:
Thank you for this. Perhaps to solve the base URL issue you could add a new setting where you can define the base URL? See how currency is added as a reference

0x416E64:
> Perhaps to solve the base URL issue you could add a new setting where you can define the base URL? 

Will try to look into that, it seems like a good option.

Donkie:
Sorry for the long response! I have a few comments:

1. Good job with the base URL, but what I intended is for it be a general base URL that could potentially be used for more than just the QR code, so you should rename the settings etc to just be "base_url" for instance.
2. The default state of the switch in the printing dialog should be off no matter the value of the base url setting, since that is the current behaviour.
3. I think it's more suited to have a dropdown with 2 options here instead of a rocker switch, since a rocker switch implies something being on or off, but in this case it's to choose between two different equally valid modes.
4. You can get rid of the help text in the printing dialog if you instead somewhere show a preview of what the QR code will contain. Then it becomes obvious to the user what happens when they change the state of the switch. Perhaps a short help text is needed to point them to the settings page where they can configure the base url.

0x416E64:
All your points make sense (and I was too focused on making *something* work in foreign codebase, 
 than think about the big picture), when I have some more free time I'll try to implement the changes.

PR #193, opened 2023-11-11
[Feature] - Add export buttons to lists

vikt-v:
Hi, 
The default print QC codes dialog  wasn't enough for me, trying to print labels using my brother label forces me to use their software.

The simplest thing was to export the data to a CSV and go from there.
Using the default export functionality of `refinedev` simply dumps the raw table data to a CSV file, this can be customized but it was enough for me.

The only issue I'm seeing is the QR code generation for those who may try to use the CSV, the code generates it as `web+spoolman:s-${spool.id}`  but it's somewhat not known to the user.

Some pictures in dark and light modes.
image
image

PR #312, opened 2024-02-11
Add export buttons for QR-Code and CSV

Sryvkver:
Hey there,
to be able to print the QR-Code with my label Printer, i needed the option to export the QR-Code as an Image.

So I added an export button to the Spool list, which allows the user to export the QR-Code as an Image.
While I was at it, I also added the option to export all the data as a CSV file.

I only now saw that there is already an open Pull Request for the CSV-Export feature (https://github.com/Donkie/Spoolman/pull/193), but my implementation is a bit different, so hopefully it is still useful, even if the other PR gets merged.
Sorry for the inconvenience to you and vikavorkin...

All translations were created with the help of DeepL, so they might not be perfect.

below some screenshots and examples:

---

## Example CSV output:
|id |comment              |location|remaining_weight|used_weight|filament.id|filament.material|filament.name|filament.vendor.name|
|---|---------------------|--------|----------------|-----------|-----------|-----------------|-------------|--------------------|
|6  |Mein toller Kommentar|TEST    |859             |141        |3          |PLA              |Test Filament|TEST                |
|7  |Mein toller Kommentar|TEST    |1000            |0          |3          |PLA              |Test Filament|TEST                |

## Example QR-Code output
(With deeplink URL and absolute URL)


thijstriemstra:
Nice.

> to be able to print the QR-Code with my label Printer,

I was also thinking about a feature that allows you to print labels straight from the spoolman interface.. Sounds like you're doing it manually now.

vikavorkin:
All good, my MR was simple to implement using the built in export functionality of refinedev, the main issue was the bug it found in refined and the time to resolve it.

This is a much better solution allowing for customization of the exported data much more granularity.

From my brief overview I have few suggestions to make it support the custom fields and few UI suggestions, I will try to find some time during the week to formalize them, if this is really please don't wait for my thoughts , I will make an issue or MR with them if need be.

client/src/components/export/functions.tsx
```diff
@@ -0,0 +1,190 @@
...
+        csvContent += "\n";
+    });
+
+    const encodedUri = encodeURI(csvContent);
+
+    const link = document.createElement("a");
+    link.setAttribute("href", encodedUri);
+    link.setAttribute("download", `${opts.filename}.csv`);
+    document.body.appendChild(link); // Required for FF
+    link.click();
```
vikavorkin, line 60: I think you should also remove the link from the body for garbage collection.

client/src/components/export/functions.tsx
```diff
@@ -0,0 +1,190 @@
...
+    for (let i = 0; i < qrCodeContents.length; i++) {
+        const content = qrCodeContents[i];
+
+        const body = JSON.stringify({
+            data: content.data,
+            box_size: opts.boxSize,
+            border: opts.padding
+        });
+
+        fetch(url, {
```
vikavorkin, line 168: You can return a the response form the API as an inline or attachment type.
Simply opening it would cause a download of the image, skipping the client side login.

client/src/components/export/dialogs/filamentExportDialog.tsx
```diff
@@ -0,0 +1,228 @@
...
+            bordered={false}
+            ghost
+            onChange={(key) => {
+              if (Array.isArray(key)) {
+                setCSVDefaultCollapse(key);
+              }
+            }}
+          >
+            <Collapse.Panel header={t("exporting.generic.csvOptions.filamentFields")} key="1">
+              <Form.Item label={t("filament.fields.id")}>
```
vikavorkin, line 70: Generating the column switches would make this code simpler and easier to maintain.
This would also allow you to reuse the logic for generation for all the export dialogs.

client/src/components/export/dialogs/filamentExportDialog.tsx
```diff
@@ -0,0 +1,228 @@
...
+            defaultActiveKey={csvDefaultCollapse}
+            bordered={false}
+            ghost
+            onChange={(key) => {
+              if (Array.isArray(key)) {
+                setCSVDefaultCollapse(key);
+              }
+            }}
+          >
+            <Collapse.Panel header={t("exporting.generic.csvOptions.filamentFields")} key="1">
```
vikavorkin, line 69: I suggest an improvement to the selection UX, instead of using switches (or checkboxes), you can use toggle-able chip items, keep them inline which make this much more compact.

Here is an example from Monday
image

client/src/components/export/dialogs/spoolExportDialog.tsx
```diff
@@ -0,0 +1,510 @@
...
+  });
+  const [qrExportOptions, setQRExportOptions] = useSavedState<QRExportOptions>("export-qrOptions-spool", {
+    boxSize: 32,
+    padding: 2,
+    useFullURL: false,
+  });
+
+  const transformSpoolToQRContent = (spool: ISpool) => {
+
+    let content = `web+spoolman:s-${spool.id}`;
```
vikavorkin, line 41: The QR code generation and parsing should be done in a common shared location,\.

client/src/components/export/functions.tsx
```diff
@@ -0,0 +1,190 @@
...
+// This is to ensure that the values are always in the same order.
+// This also ensures that id is always the first column.
+function csvSort(aAny: [string, string], bAny: [string, string]) {
+    const a = aAny[0].toString();
+    const b = bAny[0].toString();
+
+    const remainingA = a.split('.').slice(0, -1).join('.');
+    const remainingB = b.split('.').slice(0, -1).join('.');
+
+    console.log(remainingA, remainingB);
```
vikavorkin, line 126: Remove `console.log`.

client/src/components/export/dialogs/spoolExportDialog.tsx
```diff
@@ -0,0 +1,510 @@
...
+  const exportSpools = async () => {
+    switch (exportType) {
+      case 'CSV': {
+        const formValid = await csvForm.validateFields();
+
+        if (formValid.errorFields && formValid.errorFields.length > 0) {
+          return;
+        }
+
+        console.log(csvExportOptions)
```
vikavorkin, line 62: Remove.

client/src/components/export/dialogs/spoolExportDialog.tsx
```diff
@@ -0,0 +1,510 @@
...
+        </Col>
+        <Col>
+          <Form.Item label={t("exporting.generic.exportType")}>
+            <Radio.Group
+              options={[
+                { label: "QR-Code", value: "QR-Code" },
+                { label: "CSV", value: "CSV" },
+              ]}
+              onChange={(e: RadioChangeEvent) => {
+                console.log(e.target.value);
```
vikavorkin, line 108: Remove `console.log`.

spoolman/api/v1/qr.py
```diff
@@ -0,0 +1,63 @@
...
+
+class QRRequestBody(BaseModel):
+    """The request body for the QR code endpoint."""
+
+    data: Annotated[str, "The data to encode into the QR code."]
+    box_size: Annotated[int, "The size of the boxes in the QR code."] = 10
+    border: Annotated[int, "The size of the border around the QR code."] = 2
+
+
+@router.post(
```
vikavorkin, line 30: Don't see a reason have the backend generate the QR code.
You can use the exiting QRCode from antd to create a download-able image.
See https://ant.design/components/qr-code#qr-code-demo-download

Sryvkver:
@thijstriemstra 

Printing directly from the interface would be awesome, but wouldn't fit into my workflow / work with my devices.
Right now I generate all required Images, and convert them to raw bytes to send to my DIY label printer (glorified thermal printer with custom firmware)

This is mostly automated, and works well for my case. So printing directly from the interface is not something I can test with my setup.

Sryvkver:
@vikavorkin 
Thank you for all the suggestions, I will try to find some time the next days to look over them.

spoolman/api/v1/qr.py
Sryvkver: I just like to have the backend, do most of the work. Honestly I would have liked to move the CSV generation to the backend as well, but couldn't quite get it to work properly.

The main reason is, I don't know how the JS libraries will behave in different browsers / different OS's, but the backend will always work the same way.
And maybe someday we will see a native Spoolman mobile App, which would require the same generations ^^

client/src/components/export/dialogs/filamentExportDialog.tsx
Sryvkver: Good idea, always forget these things exist.

Sryvkver:
Sorry for the long delay, I had some personal matters that I needed to attend.

Now for the status update:
----
Currently most of the suggestion should be done. I will try to add the last one missing (Returning the QR-Code as an Attachment, so it gets instantly downloaded) and afterwards fix the merge conflicts.

Instead of using toggle-able Chips, I opted for a cascading Select box (see the attached Image).
This is mainly because I found the toggle-able chips from antd way to late... (Still kinda think its weird that they are under 'Data Display' and not 'Data Entry', but oh well)

Screenshot 2024-03-03 at 01 30 19

flyespresso:
Would the CSV also export the QR code values (text not image)? I think this is awesome and love to have this merged in with master. 

I've got a labeler that I can just dump a ton of data from a CSV and it'll just start printing away, including generating the QR codes from the values in the CSV.

spoolman/api/v1/qr.py
Donkie: The big downside to having the backend do this is that it opens up for other parties to start using it, and then it becomes increasingly difficult to make changes to it without breaking backwards compatibility. I would also prefer for the frontend to do this since it's fully capable of doing it. It also puts less stress on the backend server.

Styling etc can be tedious with different browsers and OSes, but that shouldn't apply to generating images using some library.

Donkie:
Please don't AI generate translations. Only translations you should add in a PR is the English ones.

CSV export seems solid 👍 

However with the QR code I have some objections.
1. The inline comment started by vikavorkin regarding backend generation
2. I feel like this could better be integrated into the printing dialog somehow? Right now you can only generate the QR codes themselves but with no text like you can with the print dialog. I feel like people will miss having that feature. I'm thinking an easy way of doing this is to have a special button in the printing dialog that exports each "item" as a picture instead of sending each page to the printer. Hopefully doable somehow in JS.

Donkie:
Closing this for now until this is worked on again

Issue #231, opened 2023-12-22
Printing Labels without QR-Codes


OrdenDesLOGO:
Would be nice to have the ability to print labels without the qr code since people with https cant use or need them

Issue #78, opened 2023-08-16
Option for manual label text with placeholders


TomW1605:
it would be great to be able to manually set the label text using placeholders. that way you could put the info you need in the format that makes most sense to the user.

failing this, can we get an option to remove the DB id from the label or at least move it to the front so you dont have the filament details with the id in the middle
ie
```
eSUN - Black
#1 - PLA+
```
would look better as
```
#1 - eSUN
Black - PLA+
```
at least in my opinion

Donkie:
I have been thinking of ways to make the layout more flexible, placeholders is a great idea. Will look into that!

TomW1605:
it would also be good to have remaining weight as an available placeholder, it would be good for stuff you dont print with often and knowing how much is there easily is more important than reprinting the label

TomW1605:
hey, have you looked at this at all? if not i might take a crack at it

Donkie:
I have not started this yet, so take a crack if you want :)

TomW1605:
i have had a quick look and i dont think it will be that hard to do.

2 things im not 100% sure on yet and would value your input.
1) what do you think is the best way to do formatting. use html tags for `<b>` and similar or use something like markup's `**`?
2) other place holder systems i have used have had a way to put extra stuff inside the placeholder so that if that value is null the extra stuff wont be included either. its useful for things like labels so they just dont show at all if the value isnt there (better than using `?`). what im not sure about here is how best to tell between the placeholder part and the other text. i dont like the idea of having a list that is checked against because a) its something to keep up to date and b) it would break if the user wanted to use that word in the other text. one idea i had was to use double `{{` for the place holders and then extra text can be between the braces (eg, `{Spool Weight: {spool_weight}g}`) but im not sold on it.

let me know what you think

im also assuming this would replace the list of checkboxes/switches that are there now with just a text field and either a link to a list of available placeholders or a collapsed section with the list. does that sound reasonable?

TomW1605:
ok, i had a shot at writing this up using html for bold and the double `{` format and it seems to work well. it is pretty much plain JS so may need some changes to work with react/TS but the basic functionality is there. the format helpers will also need some work for things like localised units.

[regex_test.txt](https://github.com/Donkie/Spoolman/files/14257271/regex_test.txt)
(sorry its a plain text file, github dosnt let you upload actual code files)

[...]

PR #314, opened 2024-02-14
add text field with placeholders for custom label format

TomW1605:
implement #78 and #231

this still needs some work.

- [ ] there are some translation strings that need to be added (i will do english, not sure about the others)
- [x] need to add some filtering to only allow a set of html tags
- [x] need to localise the number format functions (eg, use the right money symbol and `.`/`,`)
- [x] need to add a date format function
- [ ] need to add an expandable list of placeholders with examples (also want to make it so if you click one it will be inserted into the text)
- [ ] need to add explanation/examples of how the conditional prefix/suffixes work
- [ ] work out how to handle extra fields

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "spoolman-ui",
   "version": "0.16.1",
-  "lockfileVersion": 3,
+  "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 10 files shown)


Donkie:
Closing this since I have now implemented a template system myself, which will be available in the next release.
It doesn't have as many fancy formatting ideas as you had here, but that could be added later :)

Thank you for your attempt anyway!

TomW1605:
thats great.

i have had a look at your implementation and i dont think it would be hard to adapt my extra formatting options to the way you are doing it. if your open to it i will take a look at some point this weekend

Donkie:
Yes feel free

TomW1605:
i have tried a few different methods but im thinking that the easiest option is probably to use a standard markdown library such as  [react-markdown](https://github.com/remarkjs/react-markdown) for the actual formatting from the markdown to the react components. would that be ok with you?

TomW1605:
i got it working with react-markdown. there are still a few quirks i want to iron out (mostly around it not seeming to like new lines that are not full new paragraphs that take up way too much room) but i think that should be easy compared to writing from scratch

i also got the pre/post fix stuff working again, in a way that it will also be backwards compatable (ie, if you dont use it you only need a single `{}` pair)

PR #335, opened 2024-03-21
Create install_arch.sh

CooperGerman:
I had this script working on a endeavourOS machine.


Donkie:
Thank you! I see a lot of similarities between this and the debian one, perhaps the script can be split up into several scripts and then having common parts be located in some `scripts/functions/` folder or something?

CooperGerman:
Hi, yes you are right i think some common stuff can be factorized into something as mentioned. I didn't want to take the initiative of this but i can give it a try if you want.

Donkie:
Yes please do :)

CooperGerman:
It seems like the requirements.txt file needed for `install_debian.sh` is generated outside of the script itself, is this intended or can i add the  
```sh
pip install pdm
pdm export -o requirements.txt --without-hashes > requirements.txt
```
lines to the generic installer ?

Donkie:
requirements.txt is generated in the CI job (to the root dir) to prevent people from needing to install pdm, which has proved to have quite annoying dependencies on some systems. So that cannot really be changed.

CooperGerman:
OK, i understand. What is the better way for me to test the installation process in this case ? can i emulate a release in order to reuse it for an installation ?

CooperGerman:
Hi, i updated the script and updated the README accordingly. 
I only tested it on arch for now. Pre-commit was OK.

Is there a way to ensure it works without me having to setup a VM ?

scripts/install.sh
```diff
@@ -51,38 +51,63 @@ else
...
+        install_cmd="sudo $pkg_manager install -y"
+    elif [[ "$ID_LIKE" == *"arch"* ]]; then
+        pkg_manager="pacman"
+        update_cmd="sudo $pkg_manager -Sy"
+        install_cmd="sudo $pkg_manager -S --noconfirm"
+    else
+        echo -e "${ORANGE}Your operating system is not supported. Either try to install manually or reach out to spoolman github for support.${NC}"
+        exit 1
+    fi
+fi
```
Donkie, line 71: If /etc/os-release is not available it should also fail

Donkie:
> Hi, i updated the script and updated the README accordingly. I only tested it on arch for now. Pre-commit was OK.
> 
> Is there a way to ensure it works without me having to setup a VM ?

I can test it for you on debian

scripts/install.sh
```diff
@@ -51,38 +51,63 @@ else
...
-# Install needed system packages
+# Get os package manager
 #
-# Run apt-get update
-echo -e "${GREEN}Updating apt-get cache...${NC}"
-sudo apt-get update || exit 1
+if [[ -f /etc/os-release ]]; then
+    source /etc/os-release
+    if [[ "$ID_LIKE" == *"debian"* ]]; then
+        pkg_manager="apt-get"
```
Donkie, line 60: Could be nice to add an echo here saying we detected Debian-based OS, same for arch

CooperGerman: Done ;)

scripts/install.sh
```diff
@@ -51,38 +51,65 @@ else
...
 #
-# Install needed system packages
+# Get os package manager
 #
-# Run apt-get update
-echo -e "${GREEN}Updating apt-get cache...${NC}"
-sudo apt-get update || exit 1
+if [[ -f /etc/os-release ]]; then
+    source /etc/os-release
+    if [[ "$ID_LIKE" == *"debian"* ]]; then
```
Donkie, line 59: Turns out on Debian itself, ID_LIKE doesn't exist, only ID does, which is "debian". So you should check for either ID_LIKE like you do here, or `"$ID" == *"debian"*`

image

scripts/install.sh
```diff
@@ -51,38 +51,65 @@ else
...
+    elif [[ "$pkg_manager" == "pacman" ]]; then
+        packages+="python-virtualenv"
+    fi
 fi
 
 if ! command -v pip3 &>/dev/null; then
     echo -e "${ORANGE}Python pip is not installed. Installing pip...${NC}"
-    install_packages=1
+    if [[ "$pkg_manager" == "apt-get" ]]; then
+        packages+="python3-pip"
```
Donkie, line 91: You need to prepend all of these strings with a blank space, otherwise the final packages string will be like `"python3-venvpython3-piplibpq-dev"` which of course apt-get doesn't understand. Not sure how pacman handles this

CooperGerman: Of course, strange that i didn't trigger any issue with this one...

scripts/install.sh
CooperGerman: I check for both ID and ID_LIKE variables, this should do the trick for you.

Donkie:
Looks good now and seems to work fine, thank you!

PR #337, opened 2024-03-24
Adding Helm Charts For K8S Execution

Fenikkusu:
Not sure how many people might want this or not. I almost exclusively use K8S for everything, so threw this chart together (It's mostly templated, TBH) so I can start playing with Spoolman in my home cluster.

Note, as I'm new to Spoolman, I can't be certain this is 100% "ready". At the least, I know it can boot up. I've otherwise not really put any effort into this, FYI.


Donkie:
Thank you for this suggestion, however this is too specific to have in the main repo itself. Very few people will make use of this, and it clutters the repo. Additionally, I don't fully understand kubernetes and helm charts, so it will be difficult for me to maintain.

I suggest you leaving it as a branch in your own fork and then perhaps creating a discussion topic here so that people can find it if they need it.

Issue #307, opened 2024-02-08
Add vendor defaults, for example empty spool weight


sprior:
**Is your feature request related to a problem? Please describe.**
Currently every time you enter a new filament you have to fill in fields like empty spool weight with no default value

**Describe the solution you'd like**
Vendors appear to change empty spool vendors rarely (though it does happen).  If you added a field to the Vendor record "current empty spool weight" and use that as a default in the create filament screen once the vendor has been selected you would save data entry time.  When the vendor changes their spools a user would change that in the vendor screen and from that time on new filaments would default to the new value.

**Describe alternatives you've considered**
You could start to capture all the spool possibilities for a vendor, but that's probably overkill for 90% of cases.

Donkie:
Doesn't vendors have different spools for their different in-house brands? I think having this be default per vendor is too specific and would rarely be correct.
It's better to focus on having a filament database with known spool values instead.

sprior:
Every spool I ever bought from Hatchbox no matter what the color or type of plastic) was the same, Esun as well.  When I say vendor I'm talking about the in-house brand so even if it did vary by brand it would still make sense to me to have a default spool weight associated with that brand.

BigTrikerTom:
You could also define the empty coils as an independent area.
This would also allow you to store different coils for each manufacturer. Some manufacturers offer plastic, cardboard, refill and/or wooden spools. In addition, you wouldn't have to search the Internet or your own tables every time for the empty weight of the respective spool.

Just my 2 cents...

spyder007:
Right now there is an empty spool weight at the filament level, which gives you the ability to set those weights per filament.  This should allow you to account for individual differences.

If there was an 'empty spool weight' at the vendor level that "fell through" to the filament, that would be ideal.  It's essentially a "default" value for empty spool weight that is applied when you create a new filament.

I would contend, though, that the Filament shouldn't have ANY spool values on it:  that's what Spool is for.  I've created a discussion (#338) to see if this would be a desirable change, or if I am just crazy.

runtpacket:
> Doesn't vendors have different spools for their different in-house brands? I think having this be default per vendor is too specific and would rarely be correct. It's better to focus on having a filament database with known spool values instead.

I have been thinking about this lately. Could we setup a way to pull the filaments from a community ran db? Make it optional so you can either use local db or community driven one.

Donkie:
Already in progress :) https://github.com/Donkie/SpoolmanDB
Will be available in next release

[...]

Issue #299, opened 2024-02-03
Unlock ability to create spool with "higher weight" than theoretically possible


mdziekon:
**Is your feature request related to a problem? Please describe.**

Currently, when adding new spools, Spoolman always assumes that any new spool is "Filament weight" + "Empty spool weight". In theory, this is fine, but as we know it, every spool can vary slightly in weight - sometimes it's a bit less, sometimes it's a bit more. In my case, it's usually the "more" variant, usually 10, 20, sometimes 50 grams more than advertised.

Dealing with "less" works fine - you simply measure your entire spool, and put that weight into the `Measured weight` field...

...but it's impossible to insert spools with "more" weight - even if you do put the exact measured weight, the frontend will show negative number in the `Used weight` field, but more importantly, won't allow you to create / save that spool, as it doesn't pass the backend validation.

**Describe the solution you'd like**

I'd like to be able to create spools with more weight than their theoretical initial weight.

**Describe alternatives you've considered**

Right now, all I can do is put a note for each such spool, and then manually edit spool's data once I use up some filament.

MaffooClock:
Yep, [the API model specifies a constraint](https://github.com/Donkie/Spoolman/blob/a68e0d59bd18983403cc56497bdc9bca84285886/spoolman/api/v1/models.py#L188C32-L188C36) that the value should be greater than or equal to zero, which the web client violates when sending a negative value for `used_weight`.

I suppose a workaround would be to create our Filament records with a weight slightly greater than the actual weight (e.g. 1,010g for a 1Kg spool) to prevent this -- from what I've gathered (and I could be wrong), it doesn't appear that the Filament weight has any purpose other than to constrain the Spool weight.

-----

> [!NOTE]
> Secondary bug discovered

While testing this, I noticed that after receiving the HTTP 422 exception for the failed PATCH request, if you click the Save button again without changing anything, the update succeeds, and that's because the `used_weight` parameter changes back to 0 instead of using the calculated value in the Used Weight field.

I see that a new GET request is sent immediately after the failed PATCH request, the response of which is a JSON object that includes a `used_weight` property.  The Used Weight field in the UI still shows the calculated value, but the `used_weight` value retrieved in the GET request is being used instead.

dragoonmc:
I agree that this would be ideal, Filament manufacturers aren't perfect and might supply a spool that is heavy

spyder007:
[...]

Issue #292, opened 2024-01-30
Move price & weight into spool


kpolischuk:
The manufacturer of the filament I use produces spools of different weights (and prices) for same material . This forces me to create multiple records for one type of filament. Solution: Move the "price" and "weight" fields to the spools section or create possible sizes and prices for the filament provided by the manufacturer.

kpolischuk:
UPD: move available spool sizes into manufacturer section

bl4met:
Same with color, for the same reasons.

TomW1605:
recomend instead of moving it, have overrides in the spool settings.

also, it would be good if this could also be done with empty spool weight. my main supplyer has about 3 or 4 different spool designes that they use appartently at random and that all have different weights

spyder007:
I tried to encapsulate some of the issues into a discussion (#338).  It would be great to get some feedback, I'd be happy to look into implementing changes.

Donkie:
Price is available on the spool object, but weight is an article thing and that belongs to the filament object.

Issue #287, opened 2024-01-27
Load Cell Actual Measured weight


samster395:
Would be good to get functionality like [Filament Scale for octoprint](https://plugins.octoprint.org/plugins/filament_scale/) proivides where you use a HX711-based load cell to read out the current weight of the remaining filament which would then use the empty spool weight to give you the actual weight.

rufinus:
an esp32 some code and api calls to spoolman and this is done. nothing which spoolman itself has to provide.

(working an such a scale with integrated NFC reader, lets me update remaining weight if needed)

samster395:
> an esp32 some code and api calls to spoolman and this is done. nothing which spoolman itself has to provide.
> 
> (working an such a scale with integrated NFC reader, lets me update remaining weight if needed)

Surely it doesn't make sense to use another device like an esp32 when the raspberry pi already has the capability of reading the load cell, not to mention the plugin I provided is also written in python like spoolman.

I understand your point that it means there needs to be no edits to spoolman but it's not exactly the best solution.

bofh69:
Spoolman itself doesn't really do the changing of the spool's weight, it is done by the moonraker's spoolman component requesting it. It is also the component in moonraker that knows which is the active spool for the printer.

I agree with rufInus, it would be better to have it in a separate program. As you said, it doesn't make sense if you don't run spoolman on the printer's rpi, it makes even less sense if you have multiple printers connected to one Spoolman installation. Spoolman sometimes run in docker, then its a bit of a PITA to get access to the rpi's IO.

That external program can ask moonraker about the current active spool and then send the updated weight to spoolman. Both have REST APIs for it that are easy to use.

The moonraker component should perhaps then also be updated so it is possible to turn off the reporting of used filament to Spoolman, otherwise there will be a bit of double reporting. That would be a simple thing to fix.

PR #339, opened 2024-03-28
Add additional handling for spool weights

spyder007:
These are changes to implement methodology described in #338 as well as address the following feature requests:

#307 - Added `empty_spool_weight` as an optional field on vendor.  This field is used as the default value for filaments saved from this vendor.
#299 - Added `initial_weight` and `empty_weight` to the spool object.  These fields are defaulted from values on the filament, if they exist, but allow the user to set a higher initial weight than the filament + empty_weight allows.  
#292 - Adding `initial_weight` effectively moves weight to the spool level.  Users can ignore/not set weight at the filament level and then create spools with initial weights.
#287 - I added a `measure` endpoint on the spool which allows you to pass in the spool's gross weight.  Based on the initial weight of the spool,  it's empty weight, and the current usage on the spool, a new usage is calculated.

I did not modify any of the existing integration tests to ensure that old functionality would still work.  I added new tests to cover the new fields and functionality.

Note there are a few new strings.  I added to the English set, but they would need translated.


spyder007:
I added a change to add `libffi-dev` to the python-builder image... the arm7 builds were failing without it.

Donkie (changes requested):

spoolman/database/spool.py
```diff
@@ -47,11 +49,19 @@ async def create(
 ) -> models.Spool:
     """Add a new spool to the database. Leave weight empty to assume full spool."""
     filament_item = await filament.get_by_id(db, filament_id)
+
+    if empty_weight is None:
+        empty_weight = filament_item.spool_weight if filament_item.spool_weight is not None else 0
+
+    # Calculate initial_weight if not provided
+    if initial_weight is None:
+        initial_weight = (filament_item.weight if filament_item.weight is not None else 0) + empty_weight
```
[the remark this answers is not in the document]
spyder007: I modified the default values for those fields to only be set if the filament values are set.  That said, treating a null spool as zero for calculations seems to be what is done in other areas, so it is consistent.

Donkie:
Would it make sense to make this better synchronized with the existing weight values in the Filament object? The initial_weight is gross weight, while the existing filament "weight" field is net weight. Also perhaps empty_weight could be renamed to spool_weight.

spoolman/api/v1/vendor.py
```diff
@@ -33,6 +33,11 @@ class VendorParameters(BaseModel):
         description="Free text comment about this vendor.",
         example="",
     )
+    empty_spool_weight: Optional[float] = Field(
+        ge=0,
+        description="The weight of an empty spool.",
```
Donkie, line 38: Add unit

spoolman/database/spool.py
```diff
@@ -302,6 +324,68 @@ async def use_length(db: AsyncSession, spool_id: int, length: float) -> models.S
...
+
+        if initial_weight is None:
+            initial_weight = filament_info[0] + empty_weight
+
+    # Calculate the current gross weight (initial_weight - used_weight)
+    current_use = initial_weight - spool_info[1]
+
+    # if the measurement is greater than the initial weight, raise an error
+    if weight > current_use:
+        raise SpoolMeasureError("Measured weight cannot increase.")
```
Donkie, line 377: Maybe it should be able to? I don't see why not. Then you could occasionally re-measure your spools and if there has been some discrepancy such that Spoolman thinks more filament has been used than it actually has, it can be corrected.

tests_integration/tests/spool/test_measure.py
```diff
@@ -0,0 +1,144 @@
...
+        assert diff < 60
+
+        # Verify that last_used has been updated
+        diff = abs((datetime.now(tz=timezone.utc) - datetime.fromisoformat(spool["last_used"])).total_seconds())
+        assert diff < 60
+
+        current_weight = current_weight - expected_use
+
+    # Clean up
+    httpx.delete(f"{URL}/api/v1/spool/{spool['id']}").raise_for_status()
```
Donkie, line 144: You should add another test case for where the filament doesn't have the weight and used_weight fields set

Donkie (changes requested):
Added some more comments. I haven't reviewed the client part yet

spyder007:
> Would it make sense to make this better synchronized with the existing weight values in the Filament object? The initial_weight is gross weight, while the existing filament "weight" field is net weight. Also perhaps empty_weight could be renamed to spool_weight.

The issue is that the filament weight is, well, not really sensible: generally, a filament is a particular filament type.  Why would I have two different `filaments` in the system that are "Sunlu Black Matte PLA" if I have a 1KG spool and a 3KG spool.  The idea in putting the weights at the spool level was to allow filaments to be able to feed multiple spool sizes or types.  Say I respool some of my filament from a 3KG roll to a roll that is, oh, say 750 KG just because that's what I had.   Even when 1KG rolls come from the manufacturer, their gross weight can be a little more because they added some extra filament.

The general idea behind this change was to allow for that flexibility but not break existing functionality, which is why I'm grabbing the filament weight and spool weight on create.  But after initial creation, the spool should use it's own initial weight if available.

I actually toyed with writing a conversion script to go through and default all existing spools, but that felt like it would break more than it fixed.

spoolman/database/spool.py
spyder007: Right now this function utilizes the `use_weight` function, as to not replicate this functionality.  The original idea was that `measure` was just another flavor of `use` (like use weight or use length).  If I allow for the measured weight to effectively add weight to the spool, it should probably do so on it's own logical path.

If we want to allow for measurements to add, we can, but how would you want to go about this?  I'm not opposed, but I'd like to know how you would want that to look.

spoolman/database/spool.py
```diff
@@ -302,6 +324,68 @@ async def use_length(db: AsyncSession, spool_id: int, length: float) -> models.S
...
+        )
+        try:
+            filament_info = result.one()
+        except NoResultFound as exc:
+            raise ItemNotFoundError("Filament not found for spool.") from exc
+        if empty_weight is None:
+            empty_weight = filament_info[1]
+
+        if initial_weight is None:
+            initial_weight = filament_info[0] + empty_weight
```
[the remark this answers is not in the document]
spyder007: An error should occur:  in order to use measure, the initial weight or spool weight must be set.  I added additional handling for zeros and an error if the weight is not set.  

Spool weight is technically allowed to be zero and will not throw an error.

tests_integration/tests/spool/test_measure.py
spyder007: wrote a test with `random_empty_filament`

spoolman/database/spool.py
```diff
@@ -302,6 +324,68 @@ async def use_length(db: AsyncSession, spool_id: int, length: float) -> models.S
...
+    )
+
+    try:
+        spool_info = spool_result.one()
+    except NoResultFound as exc:
+        raise SpoolMeasureError("Spool not found.") from exc
+
+    initial_weight = spool_info[0]
+    empty_weight = spool_info[2]
+    if initial_weight is None or empty_weight is None:
```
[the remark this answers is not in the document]
spyder007: Yes, modified.

Donkie:
> > Would it make sense to make this better synchronized with the existing weight values in the Filament object? The initial_weight is gross weight, while the existing filament "weight" field is net weight. Also perhaps empty_weight could be renamed to spool_weight.
> 
> The issue is that the filament weight is, well, not really sensible: generally, a filament is a particular filament type. Why would I have two different `filaments` in the system that are "Sunlu Black Matte PLA" if I have a 1KG spool and a 3KG spool. The idea in putting the weights at the spool level was to allow filaments to be able to feed multiple spool sizes or types. Say I respool some of my filament from a 3KG roll to a roll that is, oh, say 750 KG just because that's what I had. Even when 1KG rolls come from the manufacturer, their gross weight can be a little more because they added some extra filament.
> 
> The general idea behind this change was to allow for that flexibility but not break existing functionality, which is why I'm grabbing the filament weight and spool weight on create. But after initial creation, the spool should use it's own initial weight if available.
> 
> I actually toyed with writing a conversion script to go through and default all existing spools, but that felt like it would break more than it fixed.

I think you misunderstood me, I don't disagree with having these two fields, but I would just want you to change so initial_weight is net weight instead of gross weight, and change so empty_weight is called spool_weight. Then it's more aligned with how the weight fields are defined in the filament object, so as to reduce confusion when using the API. Gross weight can still be calculated using initial_weight + spool_weight.

spyder007:
> > > Would it make sense to make this better synchronized with the existing weight values in the Filament object? The initial_weight is gross weight, while the existing filament "weight" field is net weight. Also perhaps empty_weight could be renamed to spool_weight.
> > 
> > 
> > The issue is that the filament weight is, well, not really sensible: generally, a filament is a particular filament type. Why would I have two different `filaments` in the system that are "Sunlu Black Matte PLA" if I have a 1KG spool and a 3KG spool. The idea in putting the weights at the spool level was to allow filaments to be able to feed multiple spool sizes or types. Say I respool some of my filament from a 3KG roll to a roll that is, oh, say 750 KG just because that's what I had. Even when 1KG rolls come from the manufacturer, their gross weight can be a little more because they added some extra filament.
> > The general idea behind this change was to allow for that flexibility but not break existing functionality, which is why I'm grabbing the filament weight and spool weight on create. But after initial creation, the spool should use it's own initial weight if available.
> > I actually toyed with writing a conversion script to go through and default all existing spools, but that felt like it would break more than it fixed.
> 
> I think you misunderstood me, I don't disagree with having these two fields, but I would just want you to change so initial_weight is net weight instead of gross weight, and change so empty_weight is called spool_weight. Then it's more aligned with how the weight fields are defined in the filament object, so as to reduce confusion when using the API. Gross weight can still be calculated using initial_weight + spool_weight.

So STORE net weight and tare weight as initial_weight and spool_weight, and only calculate gross weight as needed.  I will make those changes.

spoolman/database/spool.py
Donkie: use_weight supports negative numbers, so this restriction is completely arbitrary

migrations/versions/2024_03_26_1349-304a32906234_spool_weight_population.py
```diff
@@ -0,0 +1,62 @@
...
+        ),
+        sa.Column(
+            "spool_weight",
+            sa.Float(),
+            nullable=False,
+        ),
+    )
+
+    initial_weight = (
+        sa.select((filament.c.weight + filament.c.spool_weight).label("initial_weight"))
```
Donkie, line 47: Incorrect calculation here now that you're using net weight

client/public/locales/en/common.json
```diff
@@ -158,6 +160,8 @@
             "used_weight": "How much filament has been used from the spool. A new spool should have 0g used.",
             "remaining_weight": "How much filament is left on the spool. For a new spool this should match the spool weight.",
             "measured_weight": "How much the filament and spool weigh.",
+            "initial_weight": "The initial Weight of the filament (net weight)",
+            "spool_weight": "The weight of the spool when it is empty.",
```
Donkie, line 164: Expand this a bit explaining it will use the default weight from the vendor if not set.

client/public/locales/en/common.json
```diff
@@ -158,6 +160,8 @@
             "used_weight": "How much filament has been used from the spool. A new spool should have 0g used.",
             "remaining_weight": "How much filament is left on the spool. For a new spool this should match the spool weight.",
             "measured_weight": "How much the filament and spool weigh.",
+            "initial_weight": "The initial Weight of the filament (net weight)",
```
Donkie, line 163: Perhaps expand on this a bit as to why you would want to have this different from the filament setting. Justify to the user why adding this field was necessary.

client/public/locales/en/common.json
Donkie, line 164: Same thing here, expand a bit on when you want to change this.

Donkie:
Client-side looks solid also, nice job :) Just a few more minor comments then we can merge this.

spoolman/database/spool.py
spyder007: The issue I was running in to was what happens if your measurement is greater than the initial weight.  In this case, I just bump the initial weight to the measurement and set the used to 0.

PR #350, opened 2024-04-09
Added prometheus integration

Forest-Troll:
Added prometheus client
Added scheduler for compute current spool metrics and info

Default endpoint host:port/metrics/


spoolman/database/database.py
```diff
@@ -197,6 +209,9 @@ def schedule_tasks(scheduler: Scheduler) -> None:
         logger.info("Scheduling automatic database backup for midnight.")
         # Schedule for midnight
         scheduler.daily(datetime.time(hour=0, minute=0, second=0), _backup_task)  # type: ignore[arg-type]
+        logger.info("Scheduling automatic metric collection.")
+        logger.info("%s", datetime.time(minute=1))
```
Donkie, line 213: Remove this 2nd logger line, not useful.

spoolman/database/database.py
```diff
@@ -197,6 +209,9 @@ def schedule_tasks(scheduler: Scheduler) -> None:
...
+        scheduler.minutely(datetime.time(second=0), _metrics)
```
Donkie, line 214: This will now only get run if automatic backups are enabled and the database is a sqlite one, which is irrelevant of metrics collection.

spoolman/database/database.py
```diff
@@ -183,6 +186,15 @@ async def _backup_task() -> Optional[Path]:
     return __db.backup_and_rotate(env.get_backups_dir(), num_backups=5)
 
 
+async def _metrics() -> None:
+    """Create some useful prometheus metrics"""
+    logger.info("Start metrics collection")
+    async with get_session() as session:
+        await filament_metrics(session)
+        await spool_metrics(session)
+    logger.info("End metrics collection")
```
Donkie, line 195: These two logger lines serve better as "debug" level and not info level.

spoolman/database/database.py
```diff
@@ -183,6 +186,15 @@ async def _backup_task() -> Optional[Path]:
     return __db.backup_and_rotate(env.get_backups_dir(), num_backups=5)
 
 
+async def _metrics() -> None:
+    """Create some useful prometheus metrics"""
+    logger.info("Start metrics collection")
+    async with get_session() as session:
```
Donkie, line 192: Change this to `async for session in get_db_session():` and you don't need to create your own get_session function.

spoolman/prometheus/metrics.py
```diff
@@ -0,0 +1,71 @@
...
+                sqlalchemy.or_(
+                    models.Spool.archived.is_(False),
+                    models.Spool.archived.is_(None),
+                )
+            )
+    )
+    rows = await db.execute(stmt)
+    result = list(rows.unique().scalars().all())
+    for row in result:
+        SPOOL_PRICE.labels(str(row.id), str(row.filament_id)).set(row.price)
```
Donkie, line 50: Must handle price potentially being None

spoolman/prometheus/metrics.py
```diff
@@ -0,0 +1,71 @@
...
+    result = list(rows.unique().scalars().all())
+    for row in result:
+        FILAMENT_INFO.labels(str(row.id),
+                             row.vendor.name,
+                             row.name,
+                             row.material,
+                             row.color_hex).set(1)
+        FILAMENT_DENSITY.labels(str(row.id)).set(row.density)
+        FILAMENT_DIAMETER.labels(str(row.id)).set(row.diameter)
+        FILAMENT_WEIGHT.labels(str(row.id)).set(row.weight)
```
Donkie, line 70: Must handle row.vendor.name and row.weight potentially being None

Donkie:
Good idea!

1. You've updated all dependencies for the project, which could have unforeseen consequences, and is outside the scope of this PR. Please only add the prometheus dependencies and don't touch the others.
2. I think this should be an opt-in thing behind an environment variable, since not many people will use it and it would just sit every minute consuming resources for no reason.
3. It seems it only works if the path is `/metrics/` but not for `/metrics`, think you could fix that?
4. Please install Black to format the files correctly.

jangrewe:
Just chiming in that i'd highly appreciate this, as i use Prom/Grafana heavily to monitor my printers - and being able to integrate the filament usage would be awesome!

Forest-Troll:
> Remove this 2nd logger line, not useful.


Done

---

> This will now only get run if automatic backups are enabled and the database is a sqlite one, which is irrelevant of metrics collection.

Run with SPOOLMAN_METRICS_ENABLED=TRUE env variable only

---

> These two logger lines serve better as "debug" level and not info level.

Set debug level

---

> Change this to `async for session in get_db_session():` and you don't need to create your own get_session function.


Done

---

> Must handle price potentially being None
> Must handle row.vendor.name and row.weight potentially being None

Handles added, may be need moved to models, but i made it at metrics.py

---


> Good idea!
> 
> 1. You've updated all dependencies for the project, which could have unforeseen consequences, and is outside the scope of this PR. Please only add the prometheus dependencies and don't touch the others.
> 2. I think this should be an opt-in thing behind an environment variable, since not many people will use it and it would just sit every minute consuming resources for no reason.
> 3. It seems it only works if the path is `/metrics/` but not for `/metrics`, think you could fix that?
> 4. Please install Black to format the files correctly.


1. I am used last pdm, and it change format of lockfile. Indicated pdm version at README.
2. Added SPOOLMAN_METRICS_ENABLED, default FALSE. But you can add some application metrics by functions, if needed.
3. Change url, with some WA.
4. Installed and formated.

---

PPS: Unfortunately i'm not true developer (python and other languages). Thanks for understanding.

Donkie:
Great updates! All items resolved. Thanks for the contribution :)

PR #423, opened 2024-07-12
Optimized list views

tzahari:
I removed some information from the list views to make it easier to read.

Added right alignment for NumberColumns and price
Removed seconds from date / times
Removed decimals from spool length and weight


Donkie:
Sorry for the delay. This is good stuff, thank you for the contribution!

tzahari:
Nice project! Thanks for accepting my pull request.

PR #446, opened 2024-08-14
allow prefix and suffix with double braces in label format

TomW1605:


Donkie:
Thank you, I think it would be good to add some help somewhere on what can be achieved now. Either by modifying the existing default template or adding some help text. Right now all I see is a bunch of magic regex in the code.

TomW1605:
thats fair enough, i will add some help text and add an example to the default template

TomW1605:
i have added some examples to the default template and added an explanation to the help text. it does make the help text longer so feel free to re word it if you want

PR #458, opened 2024-08-20
Helps to identify the icon

teoteo:
When there is low contrast between color and table background

Before:
before

After:
after

After on dark background:
after_on_dark


Donkie:
Good idea, thank you!

Issue #68, opened 2023-08-13
Allow the QR code to be configured as a URL


samwiseg0:
Instead of having to open spoolman and scan the QR code. The QR code can link directly to the spool. This would allow users to be able to pull out their phone and immediately link to the spool. 

This would allow for a streamlined workflow. 

In order to not break compatibility with the current implementation this could ideally be configureable.

Donkie:
I didn't want to specify any URL in the QR code since it can change and then you'd need to re-print all your labels.

Instead the intention is that you scan the codes using the spoolman client with your phone, or a frontend can also implement scanning so that you can for example set the spool directly by showing the QR code to your printer camera. That is something that is totally possible with this setup. Fluidd should already support it.

samwiseg0:
That makes total sense but having the option would be nice. I can see where the current method should be the default .

I personally think it's a clunky workflow when I can just pull out my phone swipe up and tap. Vs having to open an app and then find the option to take a picture of the spool. 

I get that you would have to reprint your labels if your URL changes but you already have to put spoolman in front of a reverse proxy to use the camera. It is really easy to keep the url the same if anything happens in that case. A warning when using this option to the user would suffice as well as it not being the default I think would mitigate all risk.

Bloodpack:
> That makes total sense but having the option would be nice. I can see where the current method should be the default .
> 
> I personally think it's a clunky workflow when I can just pull out my phone swipe up and tap. Vs having to open an app and then find the option to take a picture of the spool.
> 
> I get that you would have to reprint your labels if your URL changes but you already have to put spoolman behind a reverse proxy to use the camera without browser hacks. It is really easy to keep the url the same if anything happens in that case. A warning when using this option to the user would suffice as well as it not being the default I think would mitigate all risk.

i agree 100% 

if you have to use a reverse proxy anyway to be able to use the camera then the URL should be specified in the QRCode, in zhat case you can change host computers ant the URL is always the same

bryanshellpuppy:
I have a QR code scanner gun, and I would like the URL feature as well. If there was a search box that I could click in, then scan the QR code that would be cool, or just a straight URL to the spool. The current setup with a camera is not ideal if you have 100+ spools and you want to scan the code and remove from stored inventory. Like default state is stored and you can click in the record to move it to in an AMS or InUse or some other term. Great work btw!

platini76:
[...]

PR #461, opened 2024-08-26
Feature: Add option to use full URL in QR codes instead of web+spoolman: prefix

samturner3:
This is a PR picked up where the [previous PR to implement this feature](https://github.com/Donkie/Spoolman/pull/311) left off.

Solves https://github.com/Donkie/Spoolman/issues/68

- Implements a `base_url` setting.
- Implements option to use base url as base for QR code scanning link. (`${baseUrl}/spool/show/{id}`)
- Includes a preview of the link on the print labels page.

If a `base_url` is not specified, it will use the window's url. (Only if 'URL' is selected; default state is still `web+spoolman:` prefix)

[Requirements](https://github.com/Donkie/Spoolman/pull/311#issuecomment-2041583504) by @Donkie 

1. be a general base URL that could potentially be used for more than just the QR code ✅ 
2. The default state of the switch in the printing dialog should be off no matter the value of the base url setting ✅ (`  const [useHTTPUrl, setUseHTTPUrl] = useSavedState("print-useHTTPUrl", false);`
3. I think it's more suited to have a dropdown with 2 options here instead of a rocker switch ✅ used radio buttons
4. You can get rid of the help text in the printing dialog if you instead somewhere show a preview of what the QR code will contain. ✅ added a preview. ("Preview: http://localhost:5173/spool/show/{id}")

## Screenshots / video

Screenshot 2024-08-26 at 6 01 28 PM

https://github.com/user-attachments/assets/354f3210-4a4d-4793-845e-a72649f63b36


Donkie:
Thank you! Well implemented, I'm glad you managed to understand the somewhat mess that is the printing frontend components.

I have no complaints, straight to merge!

PR #484, opened 2024-09-15
Prometheus metrics improve

Forest-Troll:
Renamed metrics to base units (meters, grams).
Added metric for spool initial weight.

### DIFF
--- a/spoolman/prometheus/metrics.py
+++ b/spoolman/prometheus/metrics.py
@@ -91,6 +91,5 @@ async def filament_metrics(db: AsyncSession) -> None:
             row.color_hex,
         ).set(1)
         FILAMENT_DENSITY.labels(str(row.id)).set(row.density)
-        FILAMENT_DIAMETER.labels(str(row.id)).set(row.diameter)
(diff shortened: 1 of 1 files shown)


Donkie:
Thank you for your contribution. You're correct that it should include the units in the name (and also use base units), but this is not backwards compatible, so I'm not sure if it's a good idea. What do you think?

Forest-Troll:
I think it’s normal, this not first project where metrics names changes. If we duplicate metrics it’s was provided more time series rows. And it’s was more difficulty migrate to new metrics name in future. For p8s analytical functions is possible use old and new metrics name with OR operator. You can write about this in change notes on release.

Donkie:
Sorry for late response.
I've never heard of metric names changing, I think it's an unnecessary troublemaker just for the sake of correctness. I agree the original names were not correct, but it's still not an issue understanding what they mean.

Forest-Troll:
Ok, no problem. I am revert all renaming, keeped changes at descriptions only. And added SPOOL_INITIAL_WEIGHT metric with existing stile.

Issue #493, opened 2024-09-26
Can't get Spoolman running 


SinglerGold2:
Trying to install Spoolman on a BTT 7 with CM4

while  the install procedure runs fine, I can't get it running. 
I reinstalled it, setup the OS fresh etc... nothing helps

This is what is happening when I run ./scripts/start.sh:

![image](https://github.com/user-attachments/assets/22d31653-5d9c-4f58-88d5-5e7af571042c)

My knowledge about Linux is... limited, so any help is appreciated

**Spoolman Host (please complete the following information):**
 - Installation type: Stand Alone
 - Installed version: 0.20.0 (fresh install, not an update)
 - OS and Distro:  Mainsail-OS (Debian GNU/Linux 11 (bullseye)) on a BTT Pad7 with Raspberry CM4 4gb
 - Database type: the one that comes with Spoolman

**Desktop (please complete the following information):**
 - OS: Win11
 - Browser: Edge

https://github.com/user-attachments/assets/a33efb68-bdf0-47f8-b4d1-9e7be2ef79fc

SpaceRichard:
I have the same issue, updates from 0.19.0 to 0.20.0 and it can't not start now

SinglerGold2:
I - somehow - also installed 0.19.3 and this version works....

Uro1:
The `.env.example` file in the repository which the as-installed `.env` configuration file is copied/created from has Windows CRLF line terminations instead of Unix/Linux LF line terminations, this creates the reported errors.

You can either install dos2unix to correct the files using that tool or run both `.env` and `.env-example` through `sed` and use the `file` command afterwards to verify the additional \r line terminators from CRLF have been removed:

```bash
sed -i 's/\r$//' .env .env.example
file .env && file .env.example
```

Doing this change will strip off the additional `\r` from the CRLF Windows/DOS line endings which is creating this issue and returning them to have Linux LF line endings.

After correcting the `.env` file line endings simply restart the Spoolman service with `sudo systemctl restart Spoolman` and the service will start normally and in a functional state.

I have submitted a [pull request]( https://github.com/Donkie/Spoolman/pull/497) that should fix this issue at the root cause going forwards for subsequent released, but for anyone with a broken install of 0.20.0 and unable to revert to a previos version the above instructions should get you going.


![terminal](https://github.com/user-attachments/assets/c9a29d57-2ff0-4e9a-b761-ec26f6f2b6f7)

Donkie:
[...]

PR #497, opened 2024-09-30
Update gitattributes to preserve LF line endings on the `.env.example` file

Uro1:
The `.env.example` file in the repository which the as-installed `.env` configuration file is copied/created from has Windows CRLF line terminations instead of Unix/Linux LF line terminations, this creates the reported errors in https://github.com/Donkie/Spoolman/issues/493 causing broken installed `.env` configuration files which causes the Spoolman service to enter into a non-functional state.

This `.gitattribute` change should fix this going forward for subsequent releases by ensuring the repo always maintains/converts the `.env.example` file to LF line endings.
   
For any users with a non-functional install of the v0.20.0 release resulting in Spoolman throwing out the following `.env` file errors and failing to initiate properly when checked with `systemctl status Spoolman`:
```log
klipper bash[45831]: .env: line 29: $'\r': command not found
klipper bash[45832]: .env: line 33: $'\r': command not found
klipper bash[45833]: .env: line 37: $'\r': command not found
klipper bash[45834]: .env: line 41: $'\r': command not found
klipper bash[45835]: .env: line 45: $'\r': command not found
klipper bash[45836]: .env: line 51: $'\r': command not found
klipper bash[45837]: .env: line 55: $'\r': command not found
klipper bash[45838]: .env: line 63: $'\r': command not found
klipper bash[45839]: .env: line 69: $'\r': command not found
```

Those users can manually fix their `.env` file's line endings by either installing and running `dos2unix` on the `.env` & `.env.example` files or you can run them through `sed` inside the `./Spoolman/` install directory by using the following to strip off the additional `\r` from the CRLF Windows/DOS line endings which is creating this issue and returning them to have Linux LF line endings:

```bash
sed -i 's/\r$//' .env .env.example \
file .env && file .env.example
```
By using the `file` utility users can validate that those files no longer have CRLF line endings.

After correcting the `.env` file line endings simply restart the Spoolman service with `sudo systemctl restart Spoolman` and the service will start normally and in a functional state.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,4 @@
 
 # Never modify line endings of our bash scripts
 *.sh -crlf
+.env.example text eol=lf


Donkie:
Thank you, this is a simple fix to limit my sloppyness from reaching the release again 😅

Uro1:
> Thank you, this is a simple fix to limit my sloppyness from reaching the release again 😅

No problem, I just happened to be installing Spoolman today on a fresh Klipper/Mainsail install and came across this issue myself, I saw it was a recently reported issue a few days ago, figured I'd push the fix back upstream and attach the `sed` workaround for anyone who's Spoolman may be in a non-functional state after updating to save people some headaches. 😅  👍

Issue #516, opened 2024-10-26
add archive button to spool details page


TomW1605:
**Is your feature request related to a problem? Please describe.**
when you scan a qr code on a spool label it takes you to the details page for that spool. however there is no way from there to archive a spool.  this means that despite having the qr code on the label you still have to scroll through the filament list to find a spool and archive it

**Describe the solution you'd like**
add a button to the spool details page to archive the spool

**Describe alternatives you've considered**
there isnt really any other option except scrolling through the list

**Additional context**

PR #517, opened 2024-10-26
add archive/un archive button to spool view page

TomW1605:
implements #516

### DIFF
--- a/client/src/pages/spools/show.tsx
+++ b/client/src/pages/spools/show.tsx
@@ -57,7 +57,7 @@ export const SpoolShow: React.FC<IResourceComponentsProps> = () => {
       return
     }
     // If the spool has no remaining weight, archive it immediately since it's likely not a mistake
-    if (spool.remaining_weight && spool.remaining_weight == 0) {
+    if (spool.remaining_weight != undefined && spool.remaining_weight <= 0) {
       await archiveSpool(spool, true);
     } else {
       confirm({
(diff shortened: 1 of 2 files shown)


client/src/pages/spools/show.tsx
```diff
@@ -39,6 +42,37 @@ export const SpoolShow: React.FC<IResourceComponentsProps> = () => {
...
+      invalidates: ["list", "detail"],
+    });
+  };
+
+  const archiveSpoolPopup = async (spool: ISpool|undefined) => {
+    if (spool === undefined) {
+      return
+    }
+    // If the spool has no remaining weight, archive it immediately since it's likely not a mistake
+    if (spool.remaining_weight && spool.remaining_weight == 0) {
```
Donkie, line 60: have you tested this? remaining_weight can never be both truthy and 0 at the same time. What you want to check for the first comparison here is if it's not null I guess?

Donkie:
thank you for the contribution! just a minor comment :)

client/src/pages/spools/show.tsx
TomW1605: that bit is just copied from the check done in the [table view](https://github.com/Donkie/Spoolman/blob/master/client/src/pages/spools/list.tsx#L192) so i didnt test it but yeah, it dosnt work (in either place). the result is that in both places it gives a popup confirmation even when the spool is empty. do you think it is best to just leave it as that since thats how the other button has been working or fix it in both places?

Donkie: fix it in both places would be nice :)

TomW1605: fixed both, popup will now show only if the remining weight is >0g

Donkie:
looks good, thanks for the contribution!

PR #526, opened 2024-11-01
Add Color Code Display for Single Color Filaments

Pantastisch:
Hello!

This PR enhances the user experience for tools like HueForge by displaying color codes directly beneath the color icon for single-color filaments. This addition makes it easier for users to access and utilize color information without additional steps.

Thanks for considering this improvement!

### DIFF
--- a/client/src/components/spoolIcon.css
+++ b/client/src/components/spoolIcon.css
@@ -3,6 +3,7 @@
   width: 1.5em;
   height: 2.5em;
   gap: 2px;
+  margin: 0 0.5em;
 }
 
 .spool-icon.vertical {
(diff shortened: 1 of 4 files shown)


client/src/components/spoolIcon.css
```diff
@@ -3,7 +3,6 @@
   width: 1.5em;
   height: 2.5em;
   gap: 2px;
-  margin: 0 0.5em;
```
Donkie, line 6: This doesn't look so good in the spool list, the text is butted up against the color block. Why did you remove this?

Pantastisch: Just an UI/UX improvement.

During development, I saw, that the color is not aligned to the same spot as other records. I thought this was not intended.

image

If this was your intention, I will revert the change 😊

Pantastisch: Oh okay, I see the issue! Will fix that

Donkie:
Sorry for the slow feedback, thanks for the contribution!

PR #533, opened 2024-11-09
Autofix: "PUT /api/v1/spool/45/use HTTP/1.1" 200 OK

lattaai5:
Modified the main.py file to add an environment variable SPOOLMAN_DISABLE_REQUEST_LOGGING that, when set to true, will disable console logging for requests. This should prevent the constant INFO lines from appearing in the syslog during printing operations. 
> [!CAUTION]
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
>
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.

### DIFF
--- a/spoolman/main.py
+++ b/spoolman/main.py
@@ -14,6 +14,8 @@ from prometheus_client import generate_latest
 from scheduler.asyncio.scheduler import Scheduler
 
 from spoolman import env, externaldb
+from spoolman.config import Config
+from spoolman import env, externaldb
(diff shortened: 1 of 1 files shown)


Donkie:
wtf is this

PR #578, opened 2024-12-28
Enable install script to be automated

Khaos66:
Added a script parameter to control the systemd service installation step


Donkie:
Good idea, thanks!

Issue #590, opened 2025-01-15
"Find Filaments" method in API returns all results if color not found


inukiwi:
**Describe the bug**
When searching for a filament by color through the API, I expect to get no result if there are no matches (when using a threshold of 0 and an unexisting color for example). Instead it appears that the entire parameter is ignored and all filaments are returned.

**To Reproduce**
1. Make a GET request to /api/v1/filament?color_hex={color that does exist}
2. Get a result of filaments that match the color
3. Make a GET request to /api/v1/filament?color_hex={color that does not exist}
4. Get a result of all filaments

**Expected behavior**
If there is no matching filament for the color parameter, I expect an empty result from the API.

**Spoolman Host (please complete the following information):**
 - Installation type: Docker
 - Installed version: 0.21.0
 - OS and Distro: Ubunut 22.04
 - Database type: SQLite

inukiwi:
I pulled the source to do a little debugging and located the source of this bug:
The "where in" statement in the database utils ignores empty lists, but in this case an empty list was provided as the possible filament ids (which is correct because there are no possible filaments in this case).

I created a pull request with a fix in #591

PR #591, opened 2025-01-15
Do not ignore empty lists in database selects

inukiwi:
This change resolves issue #590 .

If a where clause checks whether a value is in an empty list, it should not be ignored because it means no results should be returned.


inukiwi:
Integration tests pass with this change

Donkie:
Makes sense, I think I made it this way for a reason but I can't remember. Tests pass so should be good. Let's hope it doesn't break anything else!

PR #605, opened 2025-01-26
Add SPOOLMAN_CORS_ORIGIN variable

ecarjat:
Add the ability to have CORS origin set in a variable without having to setup the entire app in debug mode.

I have another app that's accessing Spoolman and needs CORS policy to be setup. I don't think this should need debug mode.


Donkie:
These are great, thank you. Can you add them to the .env.example file to document it? Thanks

ecarjat:
I have updated .env.example to document the variable use

PR #624, opened 2025-02-08
Bug Fix:  Ensure the QR code favicon.svg honors the base path, when configured.

bdchik:
I noticed when playing around the with the label feature that there was an empty square in the center of the QR code.  After debugging, I found that the code was referring to favicon.svg without including the base path.  This fixes that.


Donkie:
Nice find, thank you!

PR #626, opened 2025-02-12
Added adjust spool button

htsachakis:
Added adjust spool button to spools/show page for faster updating the weight/length of the spool.


Donkie:
Good idea! Just need to resolve the merge conflict and it's good to go :)

Issue #638, opened 2025-02-23
Partially patch spools, especially extra fields


tlanfer:
**Is your feature request related to a problem? Please describe.**
Currently you need to provide all extra fields if you want to update some of them.
This creates issues when i use a custom field for my home assistant integrations, but also use the happy hare spoolman integration. 
In my Home Assistant automation, i can just send all other attributes as well. The main issue here is if they have been changed since the last update, because the HA Spoolman integration updates on a schedule. But happy hare doesnt know about my custom fields, it only provides the ones it needs, which empties mine.

**Describe the solution you'd like**
It would be nice if there was a way to only provide the fields you want to change. If you want to remove a field, this could be done by explicitly sending a null value for it.
It looks like it should be easy enough to do by changing `database\spool.py`, line 234, to not replace `spool.extra`, but merge it with the given fields. I am not a python dev, i can only assume there is a smart way to change the list comprehension to do exactly that.
Maybe something like this 

```python
        elif k == "extra":
            spool.extra = [ f for f in spool.extra if f.key not in v.keys() ]
            spool.extra.append([models.SpoolField(key=k, value=v) for k, v in v.items()])
```

If i understand the code correctly, spoolman already uses a similar merge logic for other properties.

**Describe alternatives you've considered**
The alternative would be to make all clients aware of custom fields, forcing them to always GET any spools they want to update, merge on their side, and PATCH back again.

I considered if [Content-Type: application/merge-patch+json](https://datatracker.ietf.org/doc/html/rfc7386) would be a useful extension here. but seeing as other properties already merge by default, it seems sensible to also do the same for the extra fields

PR #639, opened 2025-02-24
Allow PATCH'ing partial updates of extra attributes

tlanfer:
Solves #638


tlanfer:
I had black reformat the file

Donkie:
Thanks! :)

Issue #538, opened 2024-11-12
500 Internal Server Error


BenGeba:
**Describe the bug**
I have a Home Assistant integration which queries the API every 15 minutes and uses the following route: `GET /api/v1/spool?allow_archived=False`
From time to time (in the last ~24 hours 15x) an internal server error occurs. 
The following entries can then be found in the logs:

```
INFO:     192.168.178.77:38644 - "GET /api/v1/spool?allow_archived=False HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/app/spoolman/.venv/lib/python3.11/site-packages/uvicorn/protocols/http/httptools_impl.py", line 411, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/app/spoolman/.venv/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py", line 69, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/app/spoolman/.venv/lib/python3.11/site-packages/fastapi/applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
  File "/home/app/spoolman/.venv/lib/python3.11/site-packages/starlette/applications.py", line 123, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/app/spoolman/.venv/lib/python3.11/site-packages/starlette/middleware/errors.py", line 186, in __call__
    raise exc
  File "/home/app/spoolman/.venv/lib/python3.11/site-packages/starlette/middleware/errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
  File "/home/app/spoolman/.venv/lib/python3.11/site-packages/starlette/middleware/gzip.py", line 24, in __call__
    await responder(scope, receive, send)
  File "/home/app/spoolman/.venv/lib/python3.11/site-packages/starlette/middleware/gzip.py", line 44, in __call__
    await self.app(scope, receive, self.send_with_gzip)
  File "/home/app/spoolman/.venv/lib/python3.11/site-packages/starlette/middleware/exceptions.py", line 65, in __call__
    await wrap_app_handling_exceptions(self.app, conn)(scope, receive, send)
[...]

PR #668, opened 2025-04-05
🔨 Add pool_recycle for MySQL

sinclairpaul:
I have experienced database timeout issues when running an external MariaDB database, also mentioned in #538 

I believe the issue is that the database will close idle connections as defined by the `wait_timeout` parameter within the database config.  Although I set this to the MariaDB default (28800) I still saw the timeouts after 8 hours of idle time (the connections being closed are logged at the DB server logs).

As per <https://docs.sqlalchemy.org/en/20/core/pooling.html#setting-pool-recycle> setting pool recycle seems to resolve, I've set it to 3600 as I think an hour is reasonable.

It does appear that some distributions of MariaDB set the wait_timeout to 600 seconds, although this can be changed by the user.

Been running for a few days with no issues, integration tests all pass.


Donkie:
ah nice find, please rebase with a CI fix from main

sinclairpaul:
Should be good to go

PR #694, opened 2025-05-05
Remade translation into Brazilian Portuguese

opastorello:
Remade the translation into Brazilian Portuguese and added the home text so that it could be translated, which was fixed in the home index.


KnightArch:
I'm wondering if changing the translated texts directly into the JSON files is useful because maybe the weblate PR's could overwrite it when new translations are made and merged into the master branch. I'm not very familiar with the weblate way of working things out, so I would like to point that before merging this PR. What do you think, both you @Donkie and @opastorello ? Should be the original english text the only one sent to the weblate or there is no problem translating texts directly in the project files? Can the weblate automation bot get the new texts files and figure things out into pt-BR section at the website platform?

I can translate it immediately in the weblate or just transpose the texts made by @opastorello at the weblate platform. Just let me know and please, mention me here with an @.

Thank you very much!

Donkie:
Weblate will automatically update from the source, however it's generally a bad idea to modify the source files like this directly since if someone made any change to the pt-BR language in Weblate there will be a conflict which I have to go in and fix manually.

This can be an exception since you've already done it, but in the future please use Weblate for all translating.

KnightArch:
That was my concern, @Donkie .
If there is any text that need the translation to pt-BR, please tag me with a mention. I'll get it done ASAP. I'm still getting used to weblate and its working method of getting the new text blocks from the commits of each branch. I should take some time to study it deeply and contribute more. 🤔

Thank you again for your time.

Donkie:
I think you're able to subscribe to your language in the Weblate interface, so you can see if additional translation is needed

PR #715, opened 2025-06-21
Add Spoolcount in Locations Page

JakobGue:
This adds a small display of the spool count next to each location. I added a settings entry to enable / disable the display of the number. 
image

Backstory:
While adding all my filaments into my Spoolman installation, i had to repeatedly check wether i had everything. Counting was my fastest way of checking if all the filaments of a rack where in Spoolman. To find which spool is missing where, i had to constantly count the cards in the locations, so i added this counter to save some time. Maybe some others will appreciate this small addition :)


JakobGue:
Sure, i'll change it to be always visible. Thanks for the feedback!

Donkie:
Great, thanks for the update :)

Issue #769, opened 2025-10-16
Make QR code version/modules configurable


beikeland:
**Is your feature request related to a problem? Please describe.**
I'm trying to use a 203 DPI label printer with 40x25mm labels and the default QR codes suffer from poor print quality with the current default version 3 (29 modules) QR codes leading to hard to scan QR codes. When printing custom labels with version 1 (21 modules) qr codes with, at least with the shorter form `web+spoolman:s-3` QR codes makes for better prints and more reliable scanning.

**Describe the solution you'd like**
Make QR codes configurable or default to the "coarser" version 1 / 21 modules when using the short form URL.

**Describe alternatives you've considered**
While higher resolution printer or bigger labels would also resolve this issue, most low cost label printers are limited in size and resolution.

**Additional context**

<img width="365" height="368" alt="Image" src="https://github.com/user-attachments/assets/5cc85777-d41c-4146-91bb-57ed1613cb4b" />

<img width="367" height="380" alt="Image" src="https://github.com/user-attachments/assets/94637197-c808-4df9-a320-fd53048760b3" />

beikeland:
I see there is a limit of 17 characters with 21 nodules, but this could be extended by lowering the error correction or using alphanumerical encoding.

Alphanumerical encoding would require `.toLowerCase()` in https://github.com/Donkie/Spoolman/blob/master/client/src/components/qrCodeScanner.tsx#L14 and I would suggest adding that in any case to support labels potentially made with 3rd party tools.

PR #777, opened 2025-11-04
All caps short QR codes for alphanumerical encoding

beikeland:
Related to Donkie/Spoolman#769

The current short form QR codes generates unnecessarily many data modules reducing printability on low resolution label printers. Replacing the `web+spoolman:s-` prefix with `WEB+SPOOLMAN:S-` automatically lets the QR code be encodes as Alphanumerical reducing the module count. 

The change also makes the QR code scanner match case insensitive to work with new uppercase and old lowercase QR codes as well as the long format with the full URL, so the change is backwards compatible. To further improve printability the error correction level needs to be either configurable or left at "H" for QR codes with icon and reduced to "Q" for QR codes without icons.

### DIFF
--- a/client/src/components/qrCodeScanner.tsx
+++ b/client/src/components/qrCodeScanner.tsx
@@ -13,12 +13,12 @@ const QRCodeScannerModal: React.FC = () => {
 
   const onScan = (result: string) => {
     // Check for the spoolman ID format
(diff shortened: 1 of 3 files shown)


Donkie:
Good idea, thanks for the contribution!

PR #831, opened 2026-01-27
explicitly disable pagination for useSelect

### LINKED ISSUE #827: Not enough internal filaments shown
**Describe the bug**
Not enough internal filaments are displayed when creating a spool.

**To Reproduce**
created a new spoolman.db

**Expected behavior**
See all internal Filament

**Screenshots**

Image

 - Installation type: [Docker]
 - Installed version: [ 0.23.0 (115e29b)]
 - OS and Distro: []
 - Database type: [default]

**Additional context**
Thank you for this great little helper.

nyanotech:
It seems that v5 of `@refinedev/core` changed the default pagination setting from [off](https://github.com/refinedev/refine/blob/%40refinedev/core%404.58.0/packages/core/src/definitions/helpers/handlePaginationParams/index.ts#L15) to [server](https://github.com/refinedev/refine/blob/%40refinedev/core%405.0.8/packages/core/src/definitions/helpers/handlePaginationParams/index.ts#L10), so with the version bump in spoolman `v0.23.0`, it only fetches the first 10 filaments from the server when adding a spool.

resolves #827

### DIFF
--- a/client/src/pages/filaments/create.tsx
+++ b/client/src/pages/filaments/create.tsx
@@ -64,6 +64,7 @@ export const FilamentCreate = (props: IResourceComponentsProps & CreateOrClonePr
   const { selectProps: vendorSelect } = useSelect<IVendor>({
     resource: "vendor",
     optionLabel: "name",
+    pagination: { mode: "off" },
   });
 
   const importFilament = async (filament: ExternalFilament) => {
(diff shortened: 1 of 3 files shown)


Stothed775:
When will the fix be available in a new release? Since I am not a pro docker user and don't want to break anything so I want to wait for the new release

Donkie:
I will make a new release within a couple of days.

Donkie:
this is now live in the latest release, 0.23.1

PR #844, opened 2026-02-11
Feat(printing): filament labels and .aml export

akira69:
##Summary
- Add filament label printing flow (select, template, presets)
- Add AML export (per label + per page) with configurable AML label size
- Add filament QR scheme and scanner support
- Add separate filament print presets and new print route/buttons
- Add changelog entry

##Testing
- Not run (yet)


akira69:
oops, got to test first

PR #848, opened 2026-02-12
Build `client/` inside separate Docker stage

### LINKED ISSUE #847: Build client/ dir inside Dockerfile
In order to build the Spoolman Docker image, I currently need to install Node.js, run specific commands (that I don't know off-hand).

Node `client/` app gets built automagically inside `./Dockerfile` (as a separate stage to keep it straightforward).

Options I can image:
- _leave as-is_: this makes it harder to "grab-n-go" if, for example, I'm trying to deploy via Docker on a fresh, limited device (like I had to do on my RPi just now).
- add a new stage to Dockerfile (that gets copied into the image's `/app/client/` dir): this _will_ affect docker build-times, it _will_ affect layer caching, so _order matters_ (but "correct" order depends on whether you're currently iterating on front- or back-end at the time).
- Something like a separately built or published docker image containing the client: adds complexity, not necessary, in my opinion.

**Additional context**
My current state:
- I really like Spoolman so far (started using it 1-2 weeks ago)
- Migrating to a new RPi
- Doing a "clean setup" of everything, top-to-bottom (Klipper, Docker, etc)
- Ended up going with running Docker as non-root
- ... Ran into issue: #791 
- ..... Attempt to clone the fix: PR #798
- .......  Realize it's in a fork, so clone _that_
- .......... Attempt to `docker build .` and get errors that `client/dist/` doesn't exist
- ............. Install Node on my RPi, npm-install-and-build, _then_ docker-build 😑
- ................ _hey, it's running_!!!  (thanks for helping, jpiccari!)

This saves new folks the hassle of a whole separate toolchain (install + build steps).

eib:
Attempting to fix #847

Notes:
- I _might not be building_ it correctly, yet, but it looks right as far as I can tell: just doing `npm run build`.
- Current approach: building `client/` _before_ Python, which might have implications for image layer caching.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,12 @@
+FROM node:20-slim AS client-builder
+
+# Build client
+WORKDIR /app/client
+COPY client/package*.json ./
+RUN npm ci
+COPY client/ ./
+RUN npm run build
+
 FROM python:3.14-slim-bookworm AS python-builder
 
 ENV UV_COMPILE_BYTECODE=1
@@ -51,7 +60,7 @@ RUN groupmod -g 1000 users \
     && chown -R app:app /home/app/.local/share/spoolman
 
 # Copy built client
-COPY --chown=app:app ./client/dist /home/app/spoolman/client/dist
+COPY --chown=app:app --from=client-builder /app/client/dist /home/app/spoolman/client/dist
 
 # Copy built app
 COPY --chown=app:app --from=python-builder /home/app/spoolman /home/app/spoolman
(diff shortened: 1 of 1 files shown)


eib:
On second thought, I might very well be building the client wrong. I'm currently getting this when attempting to browse (remotely, against the Docker-not-as-root-hosted spoolman-forked service):
```
Missing API URL

App was built without an API URL. Please set the VITE_APIURL environment variable to the URL of your Spoolman API.
```

Happy to fix the env var issue, but in the context of a Docker container I'm unsure where to start (at the moment).

eib:
Hmm. [CI workflow](https://github.com/Donkie/Spoolman/blob/8a29ce063a97661567121d067bd1fdf3720b89d3/.github/workflows/ci.yml#L88) indicates `VITE_APIURL` value might be `/api/v1`. ... and that appears to be working now.

Donkie:
See linked issue

PR #859, opened 2026-02-23
Remove overflow-y scroll from locations.css

VeeBack:
Since the `max-height` was removed in 251f7e2edee03720414eed50276cadb31f3c5ee2, there's no longer a reason to keep the scroll. This will remove the empty scroll bars, and make the UI neater


Donkie:

PR #862, opened 2026-02-24
feat(ui/filters): searchable filter dropdown + bulk selection + query responsiveness

akira69:
## Summary
This PR makes the filament and spool list filters easier to use on larger datasets:
- searchable filter popovers
- `Select All`, `Select None`, and `Clear` for the visible filter options
- backend search/index support to keep the affected list paths responsive

Review cleanup also carried forward the PR861 malformed-saved-state hardening into `useSavedState`, and removed split contamination that did not belong in this PR (`spoolman/import_externaldb.py` and a stray `spool_count` locale entry).

filtering
image
image

## What Changed
- Added searchable dropdown filtering to the affected table filters.
- Added bulk filter actions that operate on the currently visible searched options.
- Added backend search/query plumbing plus additive filament/spool search indexes.
- Added `/filament-name` so the UI can populate filament-name filters without fetching full filament rows.
- Guarded `useSavedState` parsing so malformed persisted state falls back to defaults and self-heals on reload.
- Removed unrelated split artifacts from this branch.

## Database Impact
- Adds Alembic index migrations:
  - `b76f1b4c3f5a` (filament/vendor search indexes)
  - `f1a3d9c2c4e1` (spool search indexes)
- Migration type: additive indexes only.
- No destructive schema or data rewrite.

## Testing Performed
- `cd client && ./node_modules/.bin/eslint src/components/column.tsx src/components/otherModels.tsx src/utils/filtering.ts src/utils/saveload.ts`
- `cd client && ./node_modules/.bin/prettier --check src/components/column.tsx src/components/otherModels.tsx src/utils/filtering.ts src/utils/saveload.ts public/locales/en/common.json`
- `uv run ruff check spoolman/api/v1/filament.py spoolman/api/v1/other.py spoolman/api/v1/spool.py spoolman/database/filament.py spoolman/database/spool.py spoolman/database/utils.py migrations/versions/2026_02_11_1700-b76f1b4c3f5a_filament_search_indexes.py migrations/versions/2026_02_11_1710-f1a3d9c2c4e1_spool_search_indexes.py`
- `cd client && npm ci && VITE_APIURL=/api/v1 npm run build`
- Local runtime: `cd /Users/dfinch/Code/Spoolman_Labels/worktrees/pr862 && PATH=.venv/bin:$PATH SPOOLMAN_DIR_DATA=/tmp/spoolman_pr_862_data uv run uvicorn spoolman.main:app --host 0.0.0.0 --port 9862`
[...]

PR #858, opened 2026-02-22
fix(custom-fields): align #773 filter/sort with current master

akira69:
## Summary
This PR is intended as the small add-on / current-master alignment layer for upstream PR #773.

Conceptually:
- `#773` is the main custom-field filter/sort feature
- `#858` is the follow-up needed to keep that work aligned with current `master`

## Important Branch Note
Although the intended scope here is the follow-up layer, this branch currently still contains the inherited `#773` custom-field commits as its base.

So reviewers should read the scope like this:
- the underlying feature comes from `#773`
- the reason this branch still matters is the follow-up alignment/fix work on top of that feature set

If the `#773` work is updated directly, this PR may no longer need to exist separately.

## What #773 Covers
The primary custom-field feature set comes from `#773`, including:
- custom-field sort/filter behavior in list tables
- backend filter/sort support for custom-field types:
  - text
  - integer / float
  - boolean
  - choice (single + multi)
  - datetime
  - integer / float ranges
- range sorting using the low-end value
- integration coverage for custom-field filter/sort behavior

## What #858 Adds
The follow-up work here is the current-master alignment layer:
- `f70d801` `fix(ts): align custom-field filter/sort typings with current master`

In practice, this fixes the frontend/state-mapping edge so custom-field sort and filter state stays attached to the correct dynamic column when `dataIndex` is array-based.

Files directly relevant to that add-on:
- `client/src/components/column.tsx`
- `client/src/utils/sorting.ts`

## Scope Boundary
This PR is only about custom/extra-field filter and sort behavior.

It does **not** include the broader searchable filter UX for standard columns. That work remains separate in `#862`.

So:
- `#858` = custom-field / extra-field filter-sort behavior and alignment
- `#862` = broader filter-dropdown UX improvements for normal columns

## Validation
- Frontend build:
  - `docker run --rm -v "$PWD:/app" -w /app/client -e VITE_APIURL=/api/v1 -e NPM_CONFIG_CACHE=/tmp/.npm -e XDG_CONFIG_HOME=/tmp/.config -e HOME=/tmp node:20 /bin/bash -lc "npm ci && npm run build"`
- Integration coverage remains tied to the `#773` custom-field stack:
  - `tests_integration/tests/fields/test_filter_sort.py`

## Out of Scope
The following remain separate:
- table interactions / resizing (`#861`)
- searchable filter dropdown UX (`#862`)
- color / hex preview UI (`#863`)
- save / clear button state styling (`#864`)
- spool-count column extraction (`#865`)
[...]

PR #865, opened 2026-02-24
feat(filaments): add spool-count column with sort/filter support

### LINKED ISSUE #201: Filaments list improvements
- Add new action for filament „Create Spool“
- Add an spool count information for every filament
- Add accordion for each filament that shows all corresponding spools if open
- Display weight in kg rather than g, hence: „1 kg“ instead of „1.000,0 g“

Thanks for this great piece of software!!

### LINKED ISSUE #207: Inventory Management
So, this will be a edge case for most people, but it  would increase the usefullness of bigger installs (like mine)

For the most standard filaments i have a 5-10 spools in stock. Currently I dont add them to spoolman. Why not? because its more trouble to select the right spool in mainsail then it is to add it to spoolman when i swap out the spool.

So like the archive flag we could have a "storage" flag. When moonraker requests the spools it by default only gets "active" spools.
Maybe later we even could extend moonraker to allow for querieng all spools. (But its not that important i think)

With the added functionality of having a list of spools by given filament (detail page) in the next release (you promised @Donkie  :-) one would have a complete inventory management. without increasing the complexity of selecting the right spool in the printer interfaces.

Having the spool count in the FilamentList itself of "active" and "storage" spools would be a great addition.

### LINKED ISSUE #539: Number of spools (and filaments) shown directly in the overview/table (as a column with a number)
It's a problem related to keeping track of all the filaments (in a fast and convenient way).

I would like to see, directly in the overview table, how many rolls of filament I have. Or how many filaments (and rolls) a manufacturer has.

It's possible to click on 'show' for a filament and then click the button 'show spools', but it feels a bit unnecessary to get the info like this. 
OR (very possible) am I missing something very obvious here..?

**Additional context**
I was a little bit confused (amazed? 😄 ) why this isn't in there yet. But wouldn't it be handy if each screen (Filaments and Manufacturers) directly show how many spools and/or filaments you have? Especially with multiple rolls of stuff it would be nice to just have a number of how many things you have of something.

For the manufacturers page show two columns with how many spools and filaments this manufacturer has.
For the filaments page show a column with how many spools you have of a certain filament.

This would give a quick and fast overview of the inventory and I assume the numbers might be quite easy to get and implement (famous last words). Just a colum with a number. Or 0 if you have none 😄 

I tried searching in the issues for the same thing but couldn't find it (please redirect if it's a duplicate), there are some things that are kind of similar but nothing 'simple' like this.

### LINKED ISSUE #688: Under Filaments should show QTY.
When viewing filaments, if you have multiple spools it should show QTY of X Filament.   

Image

If I have 4 SPools of Bambu Lab Gren I should be able to see that easily.   Say I need to know when it's down to 1 spool of that exact filament I need to order more.

### LINKED ISSUE #742: Filament Count
No issue

Being able to see a filament count next to the filament would be great.

No

**Additional context**
None

akira69:
## Summary
This PR hardens filament `spool_count` support so the filaments list can use it as a reliable inventory signal instead of a one-off display field.

## Why
The useful part of `spool_count` is not just rendering a number. It needs to stay coherent across list queries, numeric filter semantics, spool relationship changes, and adjacent PRs that touch the same query plumbing.

### Backend
- Filament list queries treat `spool_count` as a first-class sortable and filterable field.
- Filament payload hydration now keeps `spool_count` populated on list, get, and update paths instead of leaving event payloads to drift to `null` or an implied `0`.
- Distinct spool-count lookup stays numeric, including `0`, without needing empty-string semantics.
- Spool relationship changes that affect counts are wired to refresh the related filament event path.
- The `spool.filament_id` index migration remains in scope to keep count lookups responsive.

### Frontend
- The filament list exposes `spool_count` as a sortable, filterable numeric column.
- The filter path treats `0` as the real empty-state count rather than surfacing the generic `""` option.
- The column display still normalizes missing values to `0` for safety, but the backend hydration path is now intended to keep that fallback from masking stale data.

### Filament list with new `Spool Count` column visible. UI enabled to Hide/Show `Spool Count` column.
image

### Sort by `Spool Count` (asc/desc).
image

### Filter dropdown for `Spool Count` values (without `"<empty>"`)
image

## Testing Status
Targeted integration coverage was added for:
- filament get returning `spool_count`
- filament find filtering by `spool_count=0`
- filament find sorting by `spool_count`
- distinct values from `GET /api/v1/filament/spool-count`

## Testing TODO
- [ ] Run the new filament integration slice covering `spool_count` get/find/filter/sort/distinct-value cases.
- [ ] Run the frontend build.
- [ ] Manually smoke-test the filament list column, numeric filter values, and clear-filters path.

## Scope Boundary
This PR stays limited to spool-count correctness, query behavior, and integration hardening for the filament list path.
Manufacturer-level count columns and other list-query UX work remain out of scope.

## Issues
- Fixes #201
- Fixes #207
- Fixes #539
- Fixes #688
- Fixes #742

## Merge Notes
This branch still shares nearby query-plumbing touchpoints with adjacent filament-list PRs such as `#858` and `#862`, so rebase/conflict cleanup may still be needed depending on merge order.

### DIFF
--- a/client/public/locales/en/common.json
+++ b/client/public/locales/en/common.json
@@ -217,6 +217,7 @@
         "filament": "Filaments",
         "fields": {
             "id": "ID",
+            "spool_count": "Spool Count",
             "vendor_name": "Manufacturer",
             "vendor": "Manufacturer",
             "name": "Name",
--- a/client/src/components/column.tsx
+++ b/client/src/components/column.tsx
@@ -188,6 +188,7 @@ export function RichColumn<Obj extends Entity>(
 interface FilteredQueryColumnProps<Obj extends Entity> extends BaseColumnProps<Obj> {
   filterValueQuery: UseQueryResult<string[] | ColumnFilterItem[], unknown>;
   allowMultipleFilters?: boolean;
+  includeEmptyOption?: boolean;
 }
 
 export function FilteredQueryColumn<Obj extends Entity>(props: FilteredQueryColumnProps<Obj>) {
@@ -205,10 +206,12 @@ export function FilteredQueryColumn<Obj extends Entity>(props: FilteredQueryColu
       return item;
     });
   }
-  filters.push({
-    text: "<empty>",
-    value: "<empty>",
-  });
+  if (props.includeEmptyOption ?? true) {
+    filters.push({
+      text: "<empty>",
+      value: "<empty>",
+    });
+  }
 
   const typedFilters = typeFilters<Obj>(props.tableState.filters);
   const filteredValue = getFiltersForField(typedFilters, props.dataId ?? (props.id as keyof Obj));
--- a/client/src/components/column.tsx
+++ b/client/src/components/column.tsx
@@ -60,6 +60,7 @@ interface FilteredColumnProps {
   filters?: ColumnFilterItem[];
   filteredValue?: string[];
   allowMultipleFilters?: boolean;
+  includeEmptyFilter?: boolean;
   onFilterDropdownOpen?: () => void;
   loadingFilters?: boolean;
 }
--- a/client/src/pages/filaments/list.tsx
+++ b/client/src/pages/filaments/list.tsx
@@ -269,7 +269,7 @@ export const FilamentList = () => {
             ...commonProps,
             id: "spool_count",
             dataId: "spool_count",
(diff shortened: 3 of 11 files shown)


Donkie:

PR #877, opened 2026-03-06
feat(prometheus): publish spoolman_build_info metric at startup

sw1nn:
Expose version, commit hash, and build date as a Prometheus Info metric so that build metadata is available for monitoring and alerting.


Donkie:
rebased, please fix the style issue

PR #893, opened 2026-03-26
fix: address custom field review feedback for PR 773

akira69:
## Purpose

This is a helper PR for https://github.com/Donkie/Spoolman/pull/773, not an attempt to replace @daften's PR thread.

The goal here is to make the requested review fixes easy to inspect and cherry-pick.

## Commits

- 9711495 fix: address extra field filter review feedback
- 3208f8d fix: clean up custom field table state handling
- 9abd208 docs: update extra field table view description
- cb62dd6 fix: narrow boolean custom field filter syntax

## Includes

- proper circular-import cleanup for extra-field query logic
- restore empty-string filter semantics while keeping <empty> only as a UI label
- explicit 400 responses for invalid custom-field filter values
- one get_extra_fields() lookup per find() call
- corrected paginated x-total-count for extra-field-filtered queries
- expanded tests for invalid filter input and empty-value handling
- frontend sort/filter table-state cleanup and TS/lint/build fixes still missing from the live PR 773 head
- a user-facing settings copy update so the extra-fields description no longer incorrectly says extra fields cannot be sorted or filtered in table views
- boolean custom-field filter parsing narrowed to explicit true/false tokens only

## Notes

- This branch is intended to support cherry-picking into PR 773
- I posted per-thread replies on PR 773 pointing back to these commits
- I verified it locally against the same CI command paths that are red on PR 773:
  - `uv run lefthook run ci` for `style` (Ruff, format check, ESLint, Prettier)
  - `cd client && rm -f .env .env.production && echo "VITE_APIURL=/api/v1" > .env.production && npm run build` for `build-client`
- If @daften wants to cherry-pick directly, the commands are:

    git fetch https://github.com/akira69/Spoolman_Labels feat/pr773-review-feedback
    git cherry-pick 9711495 3208f8d 9abd208 cb62dd6


daften:
The commits are incorporated in #773 ;) @akira69

akira69:
773 implemented

PR #902, opened 2026-04-04
Follow up of #689 - providing spool weight deltas as part of new payload_extras attribute in Spool Event

chof747:
This PR is following up on @Donkie's comments on my previous [pull request 689](https://github.com/Donkie/Spoolman/pull/689) to provide the delta in weight on a spool update not in the payload of the SpoolEvent but as an attribute in an additional `payload_extra` entry of the event data.

It follows the recommendations and added also a test case to the integration tests to confirm the correct event propagation with the correct data in  test_spool_event.py.


spoolman/api/v1/models.py
```diff
@@ -539,6 +539,9 @@ class SpoolEvent(Event):
 
     payload: Spool = Field(description="Updated spool.")
     resource: Literal["spool"] = Field(description="Resource type.")
+    payload_extras: dict[str, float] | None = \
```
Donkie: str, float is not generic enough for a general field like this. can you rethink if this is the best way to do this?

Donkie (changes requested):

spoolman/api/v1/models.py
chof747: I have changed the value type from float to Any

Donkie: Use str instead, and then json encode all values. Thats how its solved in the extra fields feature.

spoolman/api/v1/models.py
```diff
@@ -623,6 +623,10 @@ class SpoolEvent(Event):
 
     payload: Spool = Field(description="Updated spool.")
     resource: Literal["spool"] = Field(description="Resource type.")
+    payload_extras: dict[str, Any] | None = Field(
```
Donkie: My review comment may have gotten lost. Use str instead as value, then json encode all values. Thats how it works with the extra fields feature.

Donkie (changes requested):

PR #904, opened 2026-04-05
Add filtering by extra fields in API and frontendExtra field filter

piitaya:
## Summary

Add support for filtering entities (vendors, filaments, spools) by their extra fields.

CleanShot 2026-04-05 at 18 14 53@2x

### API

The list endpoints (`GET /vendor`, `GET /filament`, `GET /spool`) now accept `extra.<key>=<value>` query parameters. Same conventions as existing filters (partial match, comma-separated OR, quoted exact match, empty for missing).

Example: `GET /api/v1/spool?extra.tag=production`

### Frontend

Extra fields of type **choice** and **boolean** now have filter dropdowns in the table columns.

### Related

Similar goal as #773 but takes a smaller, lower-risk approach (~360 lines vs ~1700). Reuses existing filter patterns with no new modules. I'm fine with closing my PR if you think we should go for #773

This PR does not include sorting by extra fields or advanced filtering (text input, number ranges, date pickers). These can be added in follow-up PRs without breaking changes.

### DIFF
--- a/client/src/components/column.tsx
+++ b/client/src/components/column.tsx
@@ -391,9 +391,11 @@ export function NumberRangeColumn<Obj extends Entity>(props: NumberColumnProps<O
 
 export function CustomFieldColumn<Obj extends Entity>(props: Omit<BaseColumnProps<Obj>, "id"> & { field: Field }) {
   const field = props.field;
+  const dataId = ("extra." + field.key) as keyof Obj & string;
   const commonProps = {
     ...props,
     id: ["extra", field.key],
+    dataId,
     title: field.name,
     sorter: false,
     transform: (value: unknown) => {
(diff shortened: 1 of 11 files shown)


paxx12:
This is great. I hope it gets merged, this would help with https://github.com/paxx12/SnapmakerU1-Extended-Firmware/pull/364!

daften:
How is this different from my PR https://github.com/Donkie/Spoolman/pull/773 ?

Donkie:
I really appreciate the effort you put in, but the way this fundamentally works, by not interpreting the json data and instead just doing a string matching, is not good enough and will likely cause a lot of confusion for users. If you'd search for "10", then it would also match fields having value "100", for example. I think @daften 's approach with #773 is the better choice.

I will make sure 773 is merged soon.

PR #908, opened 2026-04-09
Update Dockerfile with healthcheck.

meek2100:


Donkie:
is this really necessary? the backend is very simple with little to no background processes running, what can cause it to lock up?

PR #933, opened 2026-05-16
Add pyserial dependency to pyproject.toml

Phil1988:
Users with pyserial v3.5 constantly being downgraded - likely because of aiomysql -> pymysql -> pyserial constrains. This downgrade causes issues on my software, so users upgrade pyserial again to v3.5 and everything (also spoolman) is working. 

Added pyserial to the dependencies to avoid this.


Donkie:
not sure what you mean by this? Spoolman uses its own venv and doesn't itself use pyserial, so how can this be a problem?

PR #946, opened 2026-06-02
fix(ui): improve mobile responsiveness and resource search

whzy1314:
## Summary

This PR improves the Spoolman web UI on narrow/mobile viewports and adds consistent client-side resource search for the main resource list pages.

Changes include:
- Adds mobile-safe global CSS to prevent page-level horizontal overflow.
- Wraps page/header action buttons on small screens.
- Gives wide Ant Design tables their own horizontal scroll containers instead of expanding the whole document.
- Improves spacing and sizing for cards, forms, modals, drawers, pagination, scanner/media content, and footer on mobile.
- Adds reusable resource search input for Spools, Filaments, Manufacturers, and Locations.
- Adds token-based, case-insensitive fuzzy search over nested resource metadata so values like vendor, material, location, and extra fields can be found.
- Adds Vitest coverage for the resource search matching/filtering behavior.

## Motivation

The current UI is difficult to use on phones because wide table/layout content can force the full page wider than the viewport. These changes keep the document constrained to the viewport and localize horizontal scrolling to components that truly need it, such as data tables.

The resource list pages are also easier to use when users can quickly search across nested resource metadata instead of relying only on per-column filters.

## Testing

Ran locally in `client/`:

- `npm test`
- `npm run format-check`
- `npm run lint`
- `npm run build`

Also verified against a deployed test instance with a mobile/narrow browser smoke test:

- `/`
- `/spool`
- `/filament`
- `/vendor`
- `/location`

Deployed test verification included:

- Confirmed the K3s test instance is running the custom image built from this branch.
- Confirmed the live UI reports version `0.23.1 (d7e86f03fd21ef03d1f34fce98f35b7a2c4715c7)`.
- Confirmed `/spool` exposes `Search spools`.
- Confirmed searching `bambu green` filters to the `Bambu Lab - Bambu Green` spool instead of matching unrelated Bambu spools.


Donkie:

PR #948, opened 2026-06-17
fix: handle comma decimal inputs

### LINKED ISSUE #554: Incorrect handling of the decimal separator of a number when using a comma instead of a period.
**Describe the bug**
Incorrect handling of the decimal separator of a number when using a comma instead of a period.

**To Reproduce**
1. Go to any field to enter a float number.
2. Enter number with a comma ( enter 0,3 )
3. Click on another field
4. In the field where we entered 0,3, it will change to 3.0

nkgotcode:
Fixes #554.

This updates the remaining decimal `InputNumber` fields to use the existing comma-aware parser and input formatter, so values such as `0,3` are accepted consistently across filament, spool, vendor, extra-field, adjustment, and print/QR settings forms. Integer-only fields are left unchanged.

Validation:

```bash
cd client
npm run build
```

Passed:

```text
> spoolman-ui@0.23.1 build
> tsc && refine build

vite v7.3.1 building client environment for production...
✓ 6363 modules transformed.
✓ built
```

```bash
node - <<'NODE'
const fs = require('fs');
const assert = require('assert');
const dialog = fs.readFileSync('client/src/pages/printing/printingDialog.tsx', 'utf8');
assert(dialog.includes('value={previewScale}'));
assert(dialog.includes('formatter={formatNumberOnUserInput}'));
assert(dialog.includes('parser={numberParser}'));
function numberParser(value) {
  value = value?.replace(',', '.');
  value = value?.replace(/[^\d.-]/g, '');
  return parseFloat(value || '0');
}
assert.strictEqual(numberParser('0,3'), 0.3);
console.log('OK: printing preview scale InputNumber uses shared numberParser; numberParser("0,3") =>', numberParser('0,3'));
NODE
```

Passed:

```text
OK: printing preview scale InputNumber uses shared numberParser; numberParser("0,3") => 0.3
```

### DIFF
--- a/client/src/components/extraFields.tsx
+++ b/client/src/components/extraFields.tsx
@@ -3,7 +3,7 @@ import { Checkbox, Form, Input, InputNumber, Select, Typography } from "antd";
 import { FormItemProps, Rule } from "antd/es/form";
 import dayjs from "dayjs";
 import utc from "dayjs/plugin/utc";
-import { enrichText } from "../utils/parsing";
+import { enrichText, formatNumberOnUserInput, numberParserAllowEmpty } from "../utils/parsing";
 import { Field, FieldType } from "../utils/queryFields";
 import { DateTimePicker } from "./dateTimePicker";
 import { InputNumberRange } from "./inputNumberRange";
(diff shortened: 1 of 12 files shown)


Donkie:

PR #961, opened 2026-07-15
fix: ignore Kubernetes service-link value in SPOOLMAN_PORT

### LINKED ISSUE #647: Invaid Port for uvicorn on k8s
**Describe the bug**
I am getting an invalid port value when starting the container. I am using kubernetes and I had it working for a few days but I am now stuck and not sure how to specify the port for uvicorn.

**To Reproduce**
1. Start container
2. See error about uvicorn port

**Expected behavior**
Should startup and use port 80

**Screenshots**

Image

 - Installation type: Kubernetes Conatiner
 - Installed version: N/A
 - OS and Distro: Talos Linux
 - Database type: PostgreSQL

 - OS: Bazzite Linux
 - Browser Firefox

**Additional context**
Below is my deployment yaml

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: spoolman
  namespace: 3dprinting
  labels:
    app: spoolman
spec:
  replicas: 1
  progressDeadlineSeconds: 600
  revisionHistoryLimit: 2
  strategy:
    type: Recreate
  selector:
    matchLabels:
      app: spoolman
  template:
    metadata:
      labels:
        app: spoolman
    spec:
      containers:
      - name: spoolman
        image: ghcr.io/donkie/spoolman:latest
        startupProbe:
         httpGet:
           path: /
           port: 8000
         failureThreshold: 30
         periodSeconds: 10
        readinessProbe:
         httpGet:
           path: /
           port: 8000
         initialDelaySeconds: 5
         periodSeconds: 10
        livenessProbe:
         httpGet:
           path: /
           port: 8000
         initialDelaySeconds: 5
         periodSeconds: 10
        ports:
          - containerPort: 8000
        volumeMounts:
          - name: spoolman
            mountPath: "/home/app/.local/share/spoolman"
        env:
          - name: SPOOLMAN_DB_HOST
            value: "spoolman-postgresql-rw"
          - name: SPOOLMAN_DB_PORT
            value: "5432"
          - name: SPOOLMAN_DB_NAME
            value: "spoolman"
          - name: SPOOLMAN_DB_USERNAME
            value: "spoolman"
          - name: SPOOLMAN_DB_PASSWORD
            value: "spoolman"
          - name: SPOOLMAN_DB_TYPE
            value: "postgres"
      # initContainers:
      #  - name: db-perms
      #    image: busybox:latest
      #    restartPolicy: Always
      #    command: ['sh', '-c', 'chown -R 1000:1000 /home/app/.local/share/spoolman']
      #    volumeMounts:
      #      - name: spoolman
      #        mountPath: /home/app/.local/share/spoolman
      volumes:
        - name: spoolman
          persistentVolumeClaim:
            claimName: spoolman
```

lexfrei:
Kubernetes injects Docker-style service-link variables for every Service that already exists in the namespace when a pod starts. A Service named `spoolman` therefore sets `SPOOLMAN_PORT=tcp://10.43.0.1:8000` inside the container, and the entrypoint passed that straight to `uvicorn --port`:

```
Error: Invalid value for '--port': 'tcp://10.43.0.1:8000' is not a valid integer.
```

The container never comes up. It looks like it breaks at random, because it only happens when the Service is created before the pod.

The address in that value belongs to the Service, not to the container, and the port in it is the Service port — so there is nothing worth recovering from it. This ignores any `<proto>://` form (a real port never contains `://`), warns, and keeps the default port. Any other non-numeric or out-of-range value now fails with a clear message instead of an obscure uvicorn error.

## Testing

The service-link recovery path is exercised end-to-end by the existing integration matrix: the sqlite compose file now starts Spoolman with the offending value, so the container fails to boot if the entrypoint ever stops handling it.

The reject paths are covered by `tests_entrypoint.sh`, a self-contained shell test that runs the entrypoint against stub `id`/`usermod`/`uvicorn` on `PATH`. It needs neither Docker nor the image and runs as one small CI job on the runner's dash. It has teeth — against the previous entrypoint, its service-link and reject checks fail.

Verified against the published image:

```
# before
$ docker run --rm --env SPOOLMAN_PORT=tcp://10.43.0.1:8000 ghcr.io/donkie/spoolman:0.24.0
Error: Invalid value for '--port': 'tcp://10.43.0.1:8000' is not a valid integer.
$ echo $?  →  2

# after: warns, listens on 8000, /api/v1/health returns {"status":"healthy"}
```

A related collision exists for `SPOOLMAN_DB_PORT` when a Service is named `spoolman-db` — happy to send that as a follow-up if you want it.

Fixes #647.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -10,6 +10,17 @@ fail() {
     exit 1
 }
 
+# Kubernetes sets a Docker-style "service link" variable for every Service that
+# exists in the namespace when the pod starts, so a Service named "spoolman"
+# makes SPOOLMAN_PORT=tcp://<clusterIP>:8000 here. That is the Service address,
+# not a port, and uvicorn refuses to start on it. Ignore it and keep the default.
+case "$SPOOLMAN_PORT" in
+    *://*)
+        echo "Warning: ignoring SPOOLMAN_PORT=\"$SPOOLMAN_PORT\", which is a Kubernetes service link rather than a port number. Using 8000 instead." >&2
+        SPOOLMAN_PORT=8000
+        ;;
+esac
+
 if [ "$(id -u app)" -ne "$PUID" ]; then
     usermod -o -u "$PUID" app ||
         fail "Failed to update app UID to $PUID"


Donkie:
The fix looks decent, but remove anything but the change in entrypoint.sh. The testing stuff is way to overkill for this.
Try slim down the code in entrypoint.sh if possible

lexfrei:
Done. It's just entrypoint.sh now, the one case block that drops a `tcp://…` service-link back to the default port. I left the existing unquoted `exec uvicorn` line alone to keep the diff to the fix itself. #962 (the PUID/PGID one) is slimmed the same way, entrypoint only.

PR #987, opened 2026-07-31
fix(api): strip leading # from filament color codes

### LINKED ISSUE #780: Added invalid color
**Describe the bug**
I added a new filament via the REST API. There I gave it the color "#FF000000".  I got a "500: Internal Server Error." message back.

After that the GUI no longer worked, it would complain about an internal error.

The logs say the string is too long:
```
spoolman_1  |   File "/home/app/spoolman/spoolman/database/filament.py", line 268, in filament_changed
spoolman_1  |     payload=Filament.from_db(filament),
spoolman_1  |             ^^^^^^^^^^^^^^^^^^^^^^^^^^
spoolman_1  |   File "/home/app/spoolman/spoolman/api/v1/models.py", line 204, in from_db
spoolman_1  |     return Filament(
spoolman_1  |            ^^^^^^^^^
spoolman_1  |   File "/home/app/spoolman/.venv/lib/python3.11/site-packages/pydantic/main.py", line 214, in __init__
spoolman_1  |     validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
spoolman_1  |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spoolman_1  | pydantic_core._pydantic_core.ValidationError: 1 validation error for Filament
spoolman_1  | color_hex
spoolman_1  |   String should have at most 8 characters [type=string_too_long, input_value='#FF000000', input_type=str]
spoolman_1  |     For further information visit https://errors.pydantic.dev/2.10/v/string_too_long
```

**To Reproduce**

Create a vendor, then create a new filament with the following json-data:
```json
{
    "name": "PolyTerra Charcoal Black",
    "vendor_id": 1,
    "material": "PLA",
    "price": 20,
    "density": 1.24,
    "diameter": 1.75,
    "weight": 1000,
    "spool_weight": 140,
    "article_number": "PM70820",
    "comment": "",
    "settings_extruder_temp": 210,
    "settings_bed_temp": 60,
    "color_hex": "#FF000000",
    "extra": {}
}
```
Ie putting the data above in test.json and running:
```sh
wget --header="content-type: application/json" --method=POST --body-file=test.json http://localhost:7912/api/v1/filament
```

**Expected behavior**
I would expect the filament creation to fail with an error message.
I would also expect the field with the invalid data to be cleared when reading the data from the database.

 - Installation type: Docker
 - Installed version: v0.22.1
 - OS and Distro: Raspbian
 - Database type: SQLite

 - OS: Debian Bookworm
 - Browser: Firefox

Sanjays2402:
Closes #780

The `color_hex` / `multi_color_hexes` validators normalized a local copy of the value (uppercasing it and removing a leading `#`) to validate it, but returned the caller's original string. So `#FF000000` was stored with 9 characters, which no longer fits the `String(8)` `color_hex` column — the write failed and every later read of that filament 500'd, exactly as reported. Both validators now return the value with the `#` removed, matching the un-prefixed form the rest of the codebase and the frontend already assume.

Added `tests/test_color_hex_validation.py`; it fails on master (5 of 8 cases) and passes with the fix. Existing `tests/` suite: 196 passed.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/migrations/versions/2026_08_02_0900-9c1d5f2a7b31_strip_color_hex_hash.py
+++ b/migrations/versions/2026_08_02_0900-9c1d5f2a7b31_strip_color_hex_hash.py
@@ -1,0 +1,51 @@
+"""strip color hex hash.
+
+Revision ID: 9c1d5f2a7b31
+Revises: 415a8f855e14
+Create Date: 2026-08-02 09:00:00.000000
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "9c1d5f2a7b31"
+down_revision = "415a8f855e14"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    """Strip leading # characters from stored filament color codes."""
+    """Rows are rewritten one by one through the connection rather than with SQL"""
+    """string functions, so this behaves identically on sqlite, postgres, mysql"""
+    """and cockroachdb (see 304a32906234 for the cockroachdb caveat)."""
+    filament = sa.Table(
+        "filament",
+        sa.MetaData(),
+        sa.Column("id", sa.Integer, primary_key=True),
+        sa.Column("color_hex", sa.String(length=8), nullable=True),
+        sa.Column("multi_color_hexes", sa.String(length=128), nullable=True),
+    )
+
+    connection = op.get_bind()
+    rows = connection.execute(
+        sa.select(filament.c.id, filament.c.color_hex, filament.c.multi_color_hexes),
+    ).fetchall()
+
+    for row in rows:
+        values = {}
+
+        if row.color_hex is not None and "#" in row.color_hex:
+            values["color_hex"] = row.color_hex.replace("#", "")
+
+        if row.multi_color_hexes is not None and "#" in row.multi_color_hexes:
+            values["multi_color_hexes"] = row.multi_color_hexes.replace("#", "")
+
+        if values:
+            connection.execute(sa.update(filament).where(filament.c.id == row.id).values(**values))
+
+
+def downgrade() -> None:
+    """Perform the downgrade."""
+    # The leading # was never valid, so there is nothing to restore.
--- a/spoolman/api/v1/filament.py
+++ b/spoolman/api/v1/filament.py
@@ -142,7 +142,7 @@ class FilamentParameters(BaseModel):
         if len(clr) not in (6, 8):
             raise ValueError("Color code must be 6 or 8 characters long.")
 
-        return v
+        return v.removeprefix("#")
 
     @field_validator("multi_color_hexes")
     @classmethod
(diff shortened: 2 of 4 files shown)


Donkie:
Thanks, this is a good fix.

It only guards the write path though. The bad row is committed before the 500 fires, so
an instance that already hit #780 stays broken after upgrading: `GET /filament` keeps
500ing because the response model caps `color_hex` at 8 chars, and the UI can't repair
it because it can't list filaments to find the bad one. Only happens for SQLite only though, Postgres and MySQL
reject the insert outright.

Two things are necessary:

1. `Filament.from_db` in `spoolman/api/v1/models.py` should strip the `#` too, and pass
   `None` when the value still isn't a valid 6/8-char code rather than letting pydantic
   raise. Right now any junk in that column takes down the whole filament list.
2. A migration for rows already stored with a `#`, chained from head `415a8f855e14`. Do
   the rewrite row-by-row via `op.get_bind()`, not SQL string functions — it has to run
   on all four databases (precedent in `304a32906234`, including its CockroachDB note).
   `downgrade()` should be a no-op.

Migrations run unattended on everyone's server at startup, so please actually run
`python tests_integration/run.py` rather than assume it works.

Sanjays2402:
Both done in 7e4d2ee.

`Filament.from_db` now runs `color_hex`/`multi_color_hexes` through small sanitizers: strip the `#`, uppercase, and return `None` if what is left is not a valid 6/8-char code (for the multi field, `None` unless at least two valid codes survive). So a junk row degrades to a colorless filament instead of making the whole list unserializable.

Migration `9c1d5f2a7b31` chains from `415a8f855e14` and rewrites affected rows one at a time through `op.get_bind()` — select id/color_hex/multi_color_hexes, strip in Python, update only the rows that contain a `#`. No SQL string functions, same reasoning as the CockroachDB note in `304a32906234`. `downgrade()` is a no-op since the `#` was never valid.

On verification, honestly: I could not run `python tests_integration/run.py` — no working Docker daemon on this machine, and I would rather say that than claim a pass I did not see. What I did run: the unit suite (210 passed) with new cases for both sanitizers and a `from_db` test on a `#FF000000` row, RED-verified by reverting the sanitizer call. And I exercised the migration directly against SQLite — upgraded to `415a8f855e14`, inserted `color_hex='#FF0000'` plus `multi_color_hexes='#FF0000,#00FF00'` and a clean `00FF00` control, then upgraded to head:

    [(1, 'FF0000', 'FF0000,00FF00'), (2, '00FF00', None)]

then downgraded back to `415a8f855e14` cleanly. The other three engines are only covered by reading, so if CI doesn't run the integration suite on this PR I'd appreciate you kicking it off.

PR #1017, opened 2026-08-05
Use sequence-backed IDs on CockroachDB so the web UI can address them

### LINKED ISSUE #797: UI returns 404 when creating filaments with CockroachDB
**Describe the bug**
When using CockroachDB as the database backend, the Spoolman UI fails to create filaments and spools. Clicking "Save and Add" does nothing, and the API returns 404 errors. The database contains a vendor with ID `1134663890672549889`, but the UI displays and sends `1134663890672549900` in API requests, resulting in `404 Not Found` errors when trying to create filaments.

Request payload:
```json
{"name":"Jet Black","material":"PETG","vendor_id":1134663890672549900,"density":1.27,"diameter":1.75,"weight":1000,"spool_weight":193,"color_hex":"24292A","settings_extruder_temp":250,"settings_bed_temp":90,"external_id":"prusament_petg_jetblack_1000_175_p"}
```

Response:
```json
{message: "No vendor with ID 1134663890672549900 found."}
```

**To Reproduce**
1. Configure Spoolman with CockroachDB backend (`SPOOLMAN_DB_TYPE=cockroachdb`)
2. Access the Spoolman web UI
3. Navigate to create a new spool
4. Select a Filament from the dropdown (e.g., "Prusament - Jet Black (PETG, 1.75 mm, 1 kg, Plastic spool)")
7. Fill in other filament details 
8. Open the browser dev console and click "Save and Add". Nothing happens in the UI
9. See error in the dev console. API logs show `POST /api/v1/filament` returns `404 Not Found` with message `"No vendor with ID 1134663890672549900 found."`
10. Navigate to manufacturers. See ID=1134663890672549900

**Expected behavior**
- Filament and spool creation should work correctly
- The filament creation request should use the exact ID from the database
- The UI should display the correct vendor ID: `1134663890672549889`

**Screenshots**

Image

Image

Image

  - Installation type: Docker
  - Installed version: 0.22.1
  - OS and Distro: Kubernetes (k3s) on Linux
  - Database type: CockroachDB Serverless (Free Tier)

  - OS: macOS
  - Browser: Chrome

**Additional Information:**
An attempt to delete manufacturer leads to the same error.

cycsmail:
Fixes #797.

CockroachDB expands SERIAL primary keys to `DEFAULT unique_rowid()`, which produces 64-bit IDs. JavaScript numbers are only exact up to 2^53, so the web UI rounds them: the database hands out vendor ID 1134663890672549889, the UI sends back 1134663890672549900, and every follow-up request 404s.

This asks CockroachDB for sequence-backed SERIAL instead (`serial_normalization=sql_sequence_cached`), so tables are created with ordinary small IDs. It goes in as a per-connection server setting rather than a SET statement because it has to be in force when the CREATE TABLE DDL runs, and the Alembic migrations build their connection through the same `Database.connect()`.

One catch worth knowing about: `serial_normalization` only applies at table creation time, so this fixes fresh databases. An existing CockroachDB install already has `unique_rowid()` baked into its column defaults and rows with 64-bit IDs that can't be shrunk after the fact, so recreating the database is the way out there.

Added two unit tests that check the setting is passed for cockroachdb and not for the other backends; they and the existing suite pass. I don't have a live CockroachDB here, so the integration side is on the CI cockroachdb job.

### DIFF
--- a/migrations/versions/2026_09_06_1200-4e7c1b8d9f23_cockroachdb_renumber_unsafe_ids.py
+++ b/migrations/versions/2026_09_06_1200-4e7c1b8d9f23_cockroachdb_renumber_unsafe_ids.py
@@ -1,0 +1,92 @@
+"""cockroachdb renumber unsafe ids.
+
+Revision ID: 4e7c1b8d9f23
+Revises: fe4970567bb3
+Create Date: 2026-09-06 12:00:00.000000
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "4e7c1b8d9f23"
+down_revision = "fe4970567bb3"
+branch_labels = None
+depends_on = None
+
+# JavaScript numbers are only exact up to 2^53, so any ID above this gets rounded
+# by the web UI and every follow-up request 404s (#797).
+MAX_SAFE_ID = 2**53
+
+# Parent tables in FK order, with the columns in other tables that reference them.
+TABLES = (
+    ("vendor", (("filament", "vendor_id"), ("vendor_field", "vendor_id"))),
+    ("filament", (("spool", "filament_id"), ("filament_field", "filament_id"), ("tag", "filament_id"))),
+    ("spool", (("spool_field", "spool_id"), ("tag", "spool_id"))),
+)
+
+
+def upgrade() -> None:
+    """Renumber CockroachDB rows whose IDs are above 2^53 down into the safe range."""
+    """Existing CockroachDB installs got their IDs from unique_rowid(), which"""
+    """generates 64-bit values the web UI cannot represent (#797). Each unsafe row"""
+    """is copied to a new small ID, referencing rows are repointed and the old row"""
+    """is deleted, so foreign keys stay valid at every step. Only the DML lives in"""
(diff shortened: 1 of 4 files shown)


Donkie:
can't a migration fix existing tables?

cycsmail:
Yep, fair point. Added two migrations (split DML/DDL like 304a32906234 does for cockroach): the first renumbers any existing ids above 2^53 down into the safe range and repoints the FKs, the second swaps the id defaults from unique_rowid() to sequences. Also found why the cockroach CI job went red: sql_sequence_cached gives each connection its own block of ids so they stop following insertion order, which broke the vendor sort test. Switched to plain sql_sequence and verified the whole thing against a real cockroach v23.1.2, old huge-id rows and all.

cycsmail:
Rebuilt the branch on current master, my follow-up commit had dragged in a pile of unrelated master changes (that's what tripped the translations guard and the conflict), so the PR is back to just the cockroach change. Also chained the migrations after the new tags one and made the renumbering repoint tag.spool_id/filament_id as well. Left the tag ids themselves alone since that table isn't in a release yet, happy to add it if you'd rather.

PR #1050, opened 2026-08-09
Updated NO-NB norwegian language

mathiasbk:
Completed Norwegian Bokmål translation on the new V2 interface.

I have checked this with a JSON validator, and it is still valid.


Donkie:
Thanks, but please use the weblate interface instead. https://hosted.weblate.org/projects/spoolman/

PR #1112, opened 2026-08-24
feat(nfc): OpenPrintTag NDEF/CBOR codec

goeland86:
## Summary

Split out from #880, which proposed the full NFC feature (TigerTag/Qidi/OpenPrintTag
decoding, server-side reader endpoints, browser Web NFC UI, auto-create/bind) as one
large PR.

While #880 was in review, #1096 landed: a UID-only tag/scan-relay foundation, with
content decoding explicitly deferred to a later phase. @Donkie confirmed in
[this thread](https://github.com/Donkie/Spoolman/pull/880#issuecomment-5356306416)
that decoding should be phase 2 built on top of #1096's tag table and scan relay, not
a parallel `/api/v1/nfc/*` surface, and asked for the codec work split into small,
separately reviewable pieces — OpenPrintTag first, since it's what ships on retail
spools today.

This PR is that first piece: **just the OpenPrintTag codec and its unit tests.**
#880 is being closed in favor of this PR and the ones that follow it.

## What's in this PR

- `spoolman/openprinttag_codec.py` — decodes the OpenPrintTag NDEF/CBOR format from
  NFC-V (ISO 15693) tag memory: meta/main/aux CBOR sections, UUID/enum/color fields,
  and an aux-section `consumed_weight` encoder for in-place tag updates.
- `tests/test_openprinttag_codec.py` — 22 tests covering the documented field
  semantics and error paths, built against synthetic NFC-V memory dumps (capability
  container + NDEF TLV + CBOR payload) constructed in the test file itself.
- No endpoint wiring, no scan-relay integration, no hardware dependency
  (`nfcpy`/`libusb`) anywhere in the diff. `cbor2` is a dev-only dependency here since
  nothing outside the test suite imports the module yet.

## What's next — PR2/3/4 already in the pipeline

This is PR1 of four planned pieces. The other three are built and tested locally,
ready to open as their own PRs once this one's shape is agreed on:

- **PR2** — TigerTag NTAG213 binary codec + tests. Already ships on retail spools
  from Sunlu, Rosa3D, R3D and eSun, so there's a case for it following reasonably
  soon.
- **PR3** — Qidi MIFARE Classic codec + tests. Lowest priority of the three — MIFARE
  key derivation buys nothing writable today.
- **PR4** — an optional TigerTag external-database sync/lookup add-on (off by
  default, lazily imported so core never loads it unless explicitly enabled, no new
  dependency since it only uses what's already core to Spoolman).

None of the above touch #1096's endpoints either. Wiring a codec into `/tag/scan` —
decoding a scanned tag's `payload_b64`, an opt-in `create: true` flag for
auto-creating spools (default off, per @Donkie's note that a reader shouldn't
silently manufacture records), and `consumed_weight` surfaced as something a person
confirms rather than something that silently overwrites tracked weight — is a
further PR once the codec shapes here are settled.

## Verification gap — help wanted

I don't have real OpenPrintTag hardware or an official spec test-vector set, so
these tests validate internal consistency (round-trips, documented field mappings,
error paths) against my own reading of the spec, not ground truth. @Donkie suggested
test vectors from the spec repo or a dump from someone with a tagged spool as the
way to close that gap — if anyone reading this has a Prusament (or other
OpenPrintTag-tagged) spool and can pull a raw NFC-V memory dump, that would let me
add a real fixture and be a lot more confident this decodes correctly.

## Test plan

- [x] `uv run pytest tests` — 286 passed (264 pre-existing + 22 new)
- [x] `uv run ruff check .` / `ruff format --check .` — clean
- [x] `uv lock --check` — clean
- [ ] Real OpenPrintTag hardware/spec-vector validation — open, see above


Donkie:
This doesn't integrate with any of the other code, shouldn't it? My NFC PR is merged now.

goeland86:
My bad, I misunderstood what you were after. Updating the code to integrate the OpenPrintTag into your NFC work.

goeland86:
@Donkie question: do we want to test the optional spool creation from a tag by default in unit tests? Or should I leave the existing CI tests alone and add a new test stack specifically for creating spools from the tag scan if it's not matched?

Donkie:
Integration tests are sufficient if they can poke at all the important details of the code.

Is the SPOOLMAN_TAG_AUTO_CREATE_ENABLED variable necessary? I feel like as a user you're conscious about if you're using a tag reader that has the create field set on the scan request, so that gate should be a sufficient protection.

goeland86:
My understanding (possibly flawed) was that you didn't want an instance to create new spools from tags if the user hadn't specifically allowed it? Maybe I'm misunderstanding how you envisioned that feature working.

Is there a user-configurable setting that can be swapped at runtime that controls auto-spool creation from a tag that I missed?

goeland86:
Ah, ok, no I see my misunderstanding.

I read the original as a "some instance would never want spools created from a tag ever", when you only wanted to make sure that specific calls didn't auto-create a spool if it couldn't be matched after a scan. I'll remove the variable, it essentially serves as a full lock on tag->spool creation for the lifetime of the instance.

goeland86:
I ordered an NFC-V capable reader & tags to test with, but if someone is already equipped and wants to test it, please share the results. It's a good 10-14 days out at a minimum before I can test with physical hardware.

Donkie:
any update on the tests? :)

goeland86:
The hardware just got here, and I'm still trying to get my hands on a Prusament spool to get a retail-encoded tag into play.

goeland86:
Pushed an update: 
effective_material_uuid/effective_package_uuid derivation was missing (only ever came from explicit tag fields, which real tags don't carry), and several main-section fields weren't decoded at all (brand_specific_instance_id, tags, certifications, chamber temperatures, container dimensions, primary_color_lab/primary_color_ral). Both are fixed now.

Found these by decoding real OpenPrintTag dumps from Prusa's own spec repo (OpenPrintTag/openprinttag-specification, MIT licensed) — it ships real NFC-V memory dumps and ground-truth field values, which turned out to be exactly the kind of real-hardware validation this codec was missing. All core fields, and now the derived UUIDs too, match that repo's own stated ground truth exactly. Added those dumps as fixtures (tests/fixtures/openprinttag/) and wired them into both the unit tests and /tag/scan integration tests.


Still an open item: my NFC-V reader arrived, but the tags that shipped with it turned out to be MIFARE Classic (NFC-A/ISO14443), not NFC-V/ISO15693 — so genuine hardware-in-the-loop testing (reading an actual tag with the reader, not just replaying a captured dump) is still pending on getting real NFC-V tags in hand. Ideally I'll have a used prusa spool in hand at the latest September 24.

goeland86:
**Follow-up: real Prusament data doesn't fit this tag, spec analysis, root cause found, and a bug fixed.**

Tried writing the actual sample dump from Prusa's spec repo (`docs_src/sample_data/sample_tag.bin`) to the tag instead of a made-up payload. It doesn't fit, and we tracked down exactly why: the NFC-V tags I received have an NDEF capability container that claims 192 bytes (`E1 40 18 00`, MLEN=0x18 → 24×8), but each chip's own native ISO15693 "Get System Information" command — a protocol-level query independent of anything written to memory — reports the real capacity as **32 blocks × 4 bytes = 128 bytes**. The capability container overstates the chip's actual memory. This isn't a security lock (a "Get Multiple Block Security Status" query confirms no block has a lock bit set anywhere, including blocks 32+) and isn't a reader/tooling limitation (the ACR1552 is a straight pass-through here) — it's a manufacturing/formatting mismatch. Verified on a second, independent tag from the same batch (different UID) — identical CC, identical Get System Information mismatch, identical block-32 boundary — so this isn't a single defective unit, it's a batch-wide characteristic. The real Prusament dump needs 300 bytes of TLV/NDEF/CBOR, well over the tags' real 128-byte capacity.

That prompted a look at whether the real dump is over-provisioned relative to spec, independent of this batch's specific capacity issue. Pulled `main_fields.yaml` from the spec repo: `material_class` is the *only* field marked `required: true` — everything else (gtin, weights, temperatures, color, etc.) is `required: recommended` at most. `instance_uuid` turned out to be the single biggest optional cost (a raw 16-byte UUID plus CBOR overhead, ~19 bytes) — and it's the one field the spec itself says can be deduced from the tag's own UID rather than stored on-tag, which is exactly what this codec's `effective_instance_uuid` already does.

Tested five trims, using the *real* Prusament field values from the sample dump (not synthetic data), each written to the tag, read back, and decoded:

| Candidate | Fields | Size | Fit? |
|---|---|---|---|
| A | Everything except `instance_uuid` | 134B | ✗ (>128B) |
| B | A minus 5 temperature fields | 114B | ✓ |
| C | B minus `actual_netto_full_weight` | 104B | ✓ |
| D | class/type/name/brand + weight + color | 88B | ✓ |
| E | `material_class` alone | 42B | ✓ |

B through E all round-tripped correctly (PLA / Prusament / "PLA Galaxy Black" / real weight / real color). B was also validated live through a running server's `/tag/scan` endpoint.

**Bug found and fixed:** candidate A's write, truncated at the 128-byte boundary, still parsed as syntactically valid CBOR — just with the wrong type where `instance_uuid` should have been (a bare int instead of 16 bytes) — and `_parse_uuid` crashed calling `len()` on it without checking the type first. Since `tag_decode._decode_openprinttag` only catches `ValueError`, this would have surfaced as an unhandled 500 on `/tag/scan` for any genuinely torn/truncated real-world read. Fixed (commit `59367d9`): both `_parse_uuid` and `_parse_color_rgba` (same unguarded pattern) now check the CBOR-decoded type before assuming it's bytes. Added regression tests reproducing the exact truncation. Pushed through the whole branch stack.

goeland86:
At this point do you want to take it on faith that it works, given that we can use the sample payloads from the OpenPrintTag repo for the integration tests? Shortest delay to get a real Prusament spool with its tag is 24th of September.

goeland86:
**Second tag batch, live-server test, and two more fixes.**

- **New tags:** these are NXP, and `Get System Information` reports 28 blocks × 4 B = **112 B**, with a blank capability container. I wrote a valid 112-byte CC to one and re-ran the real-value trims. The 104 B, 88 B and 42 B records all round-trip byte-identical. The 114 B one doesn't fit. Prusa's un-trimmed dump still needs SLIX2-class tags (80 blocks, 320 B).
- **Live server:** bytes read back off the physical tag went through `/tag/scan`. `decoded` matched, and `external_id` matched an independently derived `uuid5(tag UID)` (UID in E0-first order). `create: true` made one correct vendor, filament and spool. Re-scans were idempotent and did not overwrite, and the websocket relay carried `decoded`.
- **Correction:** my earlier comment attributed the 500 on truncated reads to `_parse_uuid` alone. There were two paths. A zero-padded read raised `TypeError`, fixed earlier. A short, unpadded read raised `cbor2.CBORDecodeEOF`, which isn't a `ValueError`. Non-map or wrongly typed CBOR sections escaped the same way. `_decode_cbor_map` now raises `ValueError` for all of them, and both shapes return 200. Unit and integration regression tests added.
- **Also fixed:** `create: true` on a filament-linked tag created an orphan vendor, filament and spool before the link failed. It now counts as already linked.
- **Rebased** onto current `master`. Integration suite: 572/572.
- **Limits:** no retail Prusament spool has been read. Everything is spec sample data or payloads we wrote onto blank tags.

PR #1138, opened 2026-09-03
Increase external filament search limit

### LINKED ISSUE #1136: DB search not returning enough results
**Describe the bug**
When adding a spool, too few results (8) are returned from SpoolmanDB in the search dialog. The search method returns different weights for the same filament in the list, so the desired entry is often not available. Increasing the number of results is necessary in order to show the available DB entries. 

**To Reproduce**
1. Configure Spoolman to use SpoolmanDB-Community
2. From Library, Click "+ Add Spools"
3. example search: "Elegoo PLA White"
4. Realize you cannot find what you are looking for.

**Expected behavior**
Results should show all entries that fit that search. The entry I want (elegoo_pla_plawhite_1000_175_c) exists in the DB but is not shown.

 - Installation type: Docker
 - Installed version: 0.26.1
 - OS and Distro: Debian Trixie
 - Database type: SQLite

 - OS: MacOS
 - Browser Firefox

eekdood:
Increase search results to 100, consistent with the API max. Resolves #1136

### DIFF
--- a/client_v2/src/lib/api/spoolSource.ts
+++ b/client_v2/src/lib/api/spoolSource.ts
@@ -234,7 +234,7 @@ class HttpSpoolSource {
 		return filaments.map((m) => m.entity).slice(0, limit);
 	}
 
-	async searchExternalFilaments(query: string, limit = 8): Promise<ExternalFilament[]> {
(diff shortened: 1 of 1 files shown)


Donkie:
Fixed this in a better way in #1161 instead.

PR #1139, opened 2026-09-03
Expand search fields to include weight, ID, and diameter

### LINKED ISSUE #1137: Expand search scope to include more fields
Searching for spools, especially against the external DB, returns too many results.

Ability to use additional fields like weight, ID, diameter in search to narrow results further.

Increasing the number of results.

eekdood:
Expands external filament search so SpoolmanDB results can be matched by catalog id and spool weight terms like 1000g or 1kg.

- Added focused unit tests for external DB search limits, weight matching, and id matching.

Resolves #1137

### DIFF
--- a/client_v2/src/lib/api/spoolSource.ts
+++ b/client_v2/src/lib/api/spoolSource.ts
@@ -234,7 +234,7 @@ class HttpSpoolSource {
 		return filaments.map((m) => m.entity).slice(0, limit);
 	}
 
-	async searchExternalFilaments(query: string, limit = 8): Promise<ExternalFilament[]> {
+	async searchExternalFilaments(query: string, limit = 100): Promise<ExternalFilament[]> {
 		// Filtering happens server-side (/external/filament/search) so the whole catalog
 		// — thousands of entries — never has to be downloaded to the client.
 		const q = query.trim();
--- a/spoolman/api/v1/externaldb.py
+++ b/spoolman/api/v1/externaldb.py
@@ -43,7 +43,7 @@ async def search_external_filaments(
     query: Annotated[
         str,
         Query(
-            description="Search query, matched word-by-word against manufacturer, name and material.",
+            description="Search query, matched word-by-word against manufacturer, name, material, id and weight.",
             examples=["polymaker pla"],
         ),
     ],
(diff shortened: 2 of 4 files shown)


Donkie:
Thanks for taking this on, and for the tests, though it doesn't really work that well.

The search joins the fields into one string and checks every search word appears somewhere
in it. Adding more fields to that string can only make more filaments match, never fewer,
which is the opposite of what the issue asks for. Against the live catalog (6967 entries):

- `kg` on its own matches all 6967 external filaments, since the query is split on spaces and `1 kg` can
  never match as a unit, leaving `kg` sitting in every filament's text.
- `polymaker 1 kg` returns 861 and `polymaker 1kg` returns 446. Same intent, double the
  results.
- `1kg` returns 3641 where only 3522 spools are 1 kg, because `1.1kg` and `0.1kg` contain
  `1kg` as a substring.
- `100` returns 3854, of which 7 are actually 100 g spools. The rest are 1000 g.

To narrow anything, weight and diameter have to be read as numbers and compared against
the field, not matched as text. And it belongs next to the local search in
`spoolman/database/search.py`, since spools and filaments can't be searched that way
either.

eekdood:
Thanks for the feedback. I'll continue to work on it.

eekdood:
@Donkie what do you think about including some expanded search features, like field qualifiers and quoted text support? 

Handling numerical comparisons, especially with bare numbers, will require building inferences that attempt to match the appropriate search fields. The live catalog contains 94 product-name occurrences with standalone numbers, including “ABS 1.5,” “Radium Green 2.0,” and Pantone codes such as 162 and 720.

A small structured grammar would resolve this cleanly:
  - `weight:1kg`
  - `diameter:1.75`
  - `vendor:polymaker`
  - `material:pla`
  - `id:polymaker_...`
  - `"162"` to force textual matching

The convenient bare forms (100, 1kg, 1.75) could remain shorthand. This would also allow more precise searching of weights, with qualifiers like `initial_weight:` and `remaining:`

eekdood:
I’ve updated this PR to focus narrowly on numeric filtering and exact ID lookup while preserving the newly merged pagination behavior. Weight and diameter terms are now parsed and compared numerically; g/kg and mm select a specific field, while bare numbers match either field exactly without relying on inferences. IDs are handled through exact lookup rather than being added to the general text search. This applies to both external and local filament/spool search. Field qualifiers and quoted-text support will be addressed in separate PRs.

